PR #7, opened 2021-06-22
improve show method

ericphanson:
With this PR, I get

```julia
julia> long_row
Row(Schema("bar@1"), (z = [0.0 0.0 … 0.0 0.0; 0.0 0.0 … 0.0 0.0; … ; 0.0 0.0 … 0.0 0.0; 0.0 0.0 … 0.0 0.0], x = 1, y = 2))
```
where `long_row` is a row I added to the tests.

Without this tweak to the `show` method, I get:

```julia
julia> @test length(sprint(show, long_row; context = :limit => true)) < 200
Test Failed at REPL[13]:1
  Expression: length(sprint(show, long_row; context = :limit => true)) < 200
   Evaluated: 40142 < 200
ERROR: There was an error during testing
```
since it prints the whole matrix. With even bigger rows, I've had the 'ole ctrl-c -> segfault as it tries to convert the entire row into a string and the print it.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -107,4 +107,8 @@ end
     @test r === Row(Schema("bar", 1), first(Tables.rows(Arrow.Table(Arrow.tobuffer((x=[1],y=[2],z=[3]))))))
     @test r[1] === 3
     @test string(r) == "Row(Schema(\"bar@1\"), (z = 3, x = 1, y = 2))"
+
+
+    long_row = Row(Schema("bar", 1), (x=1, y=2, z=zeros(100, 100)))
+    @test length(sprint(show, long_row; context = :limit => true)) < 200
 end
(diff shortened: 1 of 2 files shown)


test/runtests.jl
```diff
@@ -107,4 +107,8 @@ end
     @test r === Row(Schema("bar", 1), first(Tables.rows(Arrow.Table(Arrow.tobuffer((x=[1],y=[2],z=[3]))))))
     @test r[1] === 3
     @test string(r) == "Row(Schema(\"bar@1\"), (z = 3, x = 1, y = 2))"
-end
\ No newline at end of file
+
+
+    long_row = Row(Schema("bar", 1), (x=1, y=2, z=zeros(100, 100)))
+    @test length(sprint(show, long_row; context = :limit => true)) < 200
```
ararslan, line 113:
```suggestion
    @test length(sprint(show, long_row; context=(:limit => true))) < 200
```

PR #11, opened 2021-06-26
add schema naming conventions

jrevels:
motivated by https://github.com/beacon-biosignals/LegolasFlux.jl/pull/1#discussion_r658936757

### DIFF
--- a/docs/src/schema.md
+++ b/docs/src/schema.md
@@ -30,7 +30,7 @@ When making deprecation/API decisions, keep in mind that multiple schema version
 
 # Naming Conventions
 
-1. Include a namespace. For example, assuming the schema is defined in `Foo.jl`, `foo.automobile` is good, `automobile` is bad.
+1. Include a namespace. For example, assuming the schema is defined in a package Foo.jl, `foo.automobile` is good, `automobile` is bad.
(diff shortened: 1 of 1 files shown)


docs/src/schema.md
```diff
@@ -28,6 +28,13 @@ It is preferable to avoid incrementing a schema's version integer ("making a bre
 
 When making deprecation/API decisions, keep in mind that multiple schema versions may be defined in the same codebase; there's nothing that prevents `@row("my-schema@1", ...)` and `@row("my-schema@2", ...)` from being defined/utilized simultaneously.
 
+# Naming Conventions
+
+1. Include a namespace. For example, assuming the schema is defined in `Foo.jl`, `foo.automobile` is good, `automobile` is bad.
```
ericphanson, line 33: I wonder if `Foo.jl` might be read as a Julia filename, like `Bar.jl/src/Foo.jl` instead of as a package name? maybe
```suggestion
1. Include a namespace. For example, assuming the schema is defined in a package Foo.jl, `foo.automobile` is good, `automobile` is bad.
```
? Since we often don't put package names in backticks.

ericphanson (approved):

PR #13, opened 2021-06-28
Be able to serialize `Legolas.Row`s as elements of table (not just entire rows)

ericphanson:
Consider this an issue rather than a PR, because I don't think the approach here really works. I think it's pretty tricky in general, and maybe generating individual methods in the macro (like @jrevels mentioned) is the way to go rather than trying to treat it in general like I do here.

Here, I just try to serialize a `Row{S}` like a `NamedTuple` implied by the schema (i.e. taking the wide type-annotations literally, i.e. if `a::Any`, then we try to make a named tuple with `a::Any`). The idea was it would be pretty simple and the types you put into the schema are literally specifying the Arrow memory layout, and you couldn't have any extra columns.

Giving up for now since I need to do other stuff, but I figured I'd push what I have since maybe at least the tour example could be re-usable in a better version of this.

### DIFF
--- a/examples/tour.jl
+++ b/examples/tour.jl
@@ -181,3 +181,52 @@ Arrow.setmetadata!(invalid, Dict("legolas_schema_qualified" => "my-child-schema@
 # given `path` supports `Base.read(path)::Vector{UInt8}` and `Base.write(path, bytes::Vector{UInt8})` then 
 # `path` will work as an argument to `Legolas.read`/`Legolas.write`. At some point, we'd like to make similar 
 # upstream improvements to Arrow.jl to render its API more path-type-agnostic.
+
+#####
+##### Using `Legolas.Row`s as elements of tables
+#####
+
+# `Legolas.Row`s may be used as elements of tables (instead of as entire rows of tables),
+# as long as they do not have any "extra" columns, that is, only the required columns given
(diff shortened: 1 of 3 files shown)


jrevels:
I might be being naive but the following seems to work and be relatively complete in terms of what it captures?

```jl
const LEGOLAS_ROW_ARROW_NAME = Symbol("JuliaLang.Legolas.Row")
Arrow.ArrowTypes.arrowname(::Type{<:Legolas.Row}) = LEGOLAS_ROW_ARROW_NAME
Arrow.ArrowTypes.ArrowType(::Type{Legolas.Row{_,F}}) where {_,F} = Tuple{String,Int,F}
Arrow.ArrowTypes.toarrow(row::Legolas.Row{S}) where {S} = (String(Legolas.schema_name(S)), Legolas.schema_version(S), getfield(row, :fields))
Arrow.ArrowTypes.JuliaType(::Val{LEGOLAS_ROW_ARROW_NAME}, ::Any) = Legolas.Row
Arrow.ArrowTypes.fromarrow(::Type{<:Legolas.Row}, name, version, fields) = Legolas.Row(Legolas.Schema(name, version), fields)
```

requires https://github.com/JuliaData/Arrow.jl/pull/229

ericphanson:
closing in favor of #16

PR #21, opened 2021-10-09
bump to Arrow.jl v2

jrevels:


src/tables.jl
```diff
@@ -84,15 +102,17 @@ Any other provided `kwargs` are forwarded to an internal invocation of `Arrow.wr
...
-    assign_to_table_metadata!(columns, (LEGOLAS_SCHEMA_QUALIFIED_METADATA_KEY => schema_qualified_string(schema),))
-    write_arrow(io_or_path, columns; kwargs...)
+function write(io_or_path, table, schema::Schema; validate::Bool=true,
+               metadata=Arrow.getmetadata(table), kwargs...)
+    validate && Legolas.validate(table, schema)
+    schema_metadata = LEGOLAS_SCHEMA_QUALIFIED_METADATA_KEY => schema_qualified_string(schema)
+    if isnothing(metadata)
+        metadata = (schema_metadata,)
+    else
+        metadata = Set(collect(metadata))
```
ericphanson, line 112: I think `collect` shouldn't be necessary: `Arrow.getmetadata` returns an `AbstractDict` which iterates pairs, and `Arrow.write`'s `metadata` keyword argument says the input must interate pairs. And `Set`'s docstring says it accepts an iterable to create the set.
```suggestion
        metadata = Set(metadata)
```

src/tables.jl
```diff
@@ -84,15 +102,17 @@ Any other provided `kwargs` are forwarded to an internal invocation of `Arrow.wr
...
+        push!(metadata, schema_metadata)
```
ericphanson, line 113: CodeCov says we're missing a test for this

ericphanson (commented):
Exciting! Looks good to me except for a missing test for `Legolas.write` with metadata (passed-in or pre-existing).

PR #28, opened 2021-11-16
more descriptive error when unknown Legolas.Schema is encountered

jrevels:


src/rows.jl
```diff
@@ -84,14 +84,35 @@ Return the `Legolas.Schema` instance that corresponds to the parent of the given
 @inline schema_parent(::Type{<:Schema}) = nothing
 @inline schema_parent(schema::Schema) = schema_parent(typeof(schema))
 
-"""
-    schema_qualified_string(::Legolas.Schema{name,version})
+Base.show(io::IO, schema::Schema) = print(io, "Schema(\"$(schema_name(schema))@$(schema_version(schema))\")")
```
ericphanson, line 87: Is this needed? I get the printing being the same without this:
```julia
julia> using Legolas

julia> Legolas.Schema("hi@1")
Schema("hi@1")

julia> using Legolas: Schema, schema_version, schema_name

julia> Base.show(io::IO, schema::Schema) = print(io, "Schema(\"$(schema_name(schema))@$(schema_version(schema))\")")

julia> Legolas.Schema("hi@1")
Schema("hi@1")
```
```suggestion
```

ericphanson (approved):
Nice, looks helpful!

src/rows.jl
jrevels: This wasn't added by this PR - this line just got moved since I split up this code regions into 2 regions for the sake of the PR, so it shows up in the diff twice (once `+` here, and `-` below from where it was moved)

ericphanson: Ohh, got it, thanks

PR #29, opened 2021-12-11
Add `Schema` methods for `Row`s and make `write` more permissive

ararslan:
If you need to `Legolas.write` and all you have is the row type definition you get from `@row`, you need to extract the schema in order to pass it to `Legolas.write`. However, there's no particularly convenient or obvious way to do this. To make this and other such situations easier, I've made the following changes:

- `Schema` constructor methods for `Row`, `Type{<:Row}`, `Schema`, and `Type{<:Schema}`. This allows you to call `Schema` on something which has or is itself a schema and get back said schema.
- Remove the type restriction for the schema argument of `Legolas.write`, thereby permitting passing a row, row type, schema, or schema type (or anything else for which a `Schema` method exists).

### DIFF
--- a/src/rows.jl
+++ b/src/rows.jl
@@ -57,6 +57,9 @@ function Schema(s::AbstractString)
     throw(ArgumentError("argument is not a valid `Legolas.Schema` string: \"$s\""))
 end
 
+Schema(::Type{S}) where {S<:Schema} = S()
+Schema(schema::Schema) = schema
+
 """
     schema_name(::Type{<:Legolas.Schema{name}})
     schema_name(::Legolas.Schema{name})
@@ -198,6 +201,16 @@ Row{S}(args...; kwargs...) where {S} = Row(S(), args...; kwargs...)
 Row(schema::Schema, fields) = Row(schema, NamedTuple(Tables.Row(fields)))
 Row(schema::Schema, fields::Row) = Row(schema, getfield(fields, :fields))
 Row(schema::Schema, fields::NamedTuple) = Row(schema; fields...)
+
+"""
+    Legolas.Schema(row::Row)
+    Legolas.Schema(::Type{<:Row})
+
+Return the schema for the given [`Legolas.Row`](@ref) or row type.
+"""
+Schema(row::Row) = Schema(typeof(row))
+Schema(::Type{Row{S}}) where {S} = S()
+Schema(::Type{Row{S,F}}) where {S,F} = S()
 
 Base.propertynames(row::Row) = propertynames(getfield(row, :fields))
 Base.getproperty(row::Row, name::Symbol) = getproperty(getfield(row, :fields), name)
--- a/src/tables.jl
+++ b/src/tables.jl
@@ -91,10 +91,13 @@ function read(io_or_path; validate::Bool=true)
 end
 
 """
-    Legolas.write(io_or_path, table, schema::Schema; validate::Bool=true, kwargs...)
+    Legolas.write(io_or_path, table, row_or_schema; validate::Bool=true, kwargs...)
 
 Write `table` to `io_or_path`, inserting the appropriate `$LEGOLAS_SCHEMA_QUALIFIED_METADATA_KEY`
 field in the written out Arrow metadata.
+
+`row_or_schema` can be a [`Legolas.Row`](@ref) object or type, or a [`Legolas.Schema`](@ref)
+object or type.
 
 If `validate` is `true`, `Legolas.validate` will be called on the table before it written out.
 
@@ -102,8 +105,9 @@ Any other provided `kwargs` are forwarded to an internal invocation of `Arrow.wr
 
 Note that `io_or_path` may be any type that supports `Base.write(io_or_path, bytes::Vector{UInt8})`.
 """
-function write(io_or_path, table, schema::Schema; validate::Bool=true,
+function write(io_or_path, table, row_or_schema; validate::Bool=true,
                metadata=Arrow.getmetadata(table), kwargs...)
+    schema = Schema(row_or_schema)
     validate && Legolas.validate(table, schema)
     schema_metadata = LEGOLAS_SCHEMA_QUALIFIED_METADATA_KEY => schema_qualified_string(schema)
     if isnothing(metadata)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -92,11 +92,13 @@ end
     path = MyPath(joinpath(root, "baz.arrow"))
     Baz = @row("baz@1", a, b)
     t = [Baz(a=1, b=2), Baz(a=3, b=4)]
-    Legolas.write(path, t, Schema("baz", 1))
-    @test t == Baz.(Tables.rows(Legolas.read(path)))
-    tbl = Arrow.Table(Legolas.tobuffer(t, Schema("baz", 1); metadata=("a" => "b", "c" => "d")))
-    @test Set(Arrow.getmetadata(tbl)) == Set((Legolas.LEGOLAS_SCHEMA_QUALIFIED_METADATA_KEY => "baz@1",
-                                              "a" => "b", "c" => "d"))
+    for arg in [Baz, Baz(), Schema(Baz), Schema{:baz,1}]
+        Legolas.write(path, t, arg)
+        @test t == Baz.(Tables.rows(Legolas.read(path)))
+        tbl = Arrow.Table(Legolas.tobuffer(t, Schema("baz", 1); metadata=("a" => "b", "c" => "d")))
+        @test Set(Arrow.getmetadata(tbl)) == Set((Legolas.LEGOLAS_SCHEMA_QUALIFIED_METADATA_KEY => "baz@1",
+                                                  "a" => "b", "c" => "d"))
+    end
 
     struct Foo
         meta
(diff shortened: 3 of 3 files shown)


src/tables.jl
```diff
@@ -91,19 +91,23 @@ function read(io_or_path; validate::Bool=true)
 end
 
 """
-    Legolas.write(io_or_path, table, schema::Schema; validate::Bool=true, kwargs...)
+    Legolas.write(io_or_path, table, row_or_schema; validate::Bool=true, kwargs...)
```
jrevels, line 94: Let’s not do this and instead keep the input domain simple for callers; this function really only needs the schema to do its job, not a row, so we should only have it accept a schema

Rather the “right” thing to do is to solve pain points with extracting/constructing schemas themselves (which the rest of this PR is adding mechanisms for)

src/rows.jl
```diff
@@ -199,6 +202,16 @@ Row(schema::Schema, fields) = Row(schema, NamedTuple(Tables.Row(fields)))
 Row(schema::Schema, fields::Row) = Row(schema, getfield(fields, :fields))
 Row(schema::Schema, fields::NamedTuple) = Row(schema; fields...)
 
+"""
+    Legolas.Schema(row::Row)
+    Legolas.Schema(::Type{<:Row})
+
+Return the schema for the given [`Legolas.Row`](@ref) or row type.
+"""
+Schema(row::Row) = Schema(typeof(row))
```
jrevels, line 211: I’m mostly fine with this but I wonder if it should be a separate function rather than a constructor

the reason being that it encourages writing code that naive authors might think is “generic” but isn’t as generic as most producer/consumer code probably should be (ie not caring whether you’re given an actual Legolas.Row or just a compliant Tables.jl row)

actually, maybe that brings up a good point - can you point me to code where this mechanism is useful in practice? I wonder whether the need here is actually pointing out an upstream code smell or something (since it implies the existence of code that “cares” if it’s being given a Legolas.Row, without that code itself being the source of the Legolas.Row)

(keep in mind this is OSS, so if internal Beacon code, can link on Slack)

ararslan: It's internal code but the gist is that I often find myself writing `MyRowType = @row(...)`, applying `MyRowType` to some rows, then wanting to `write` out the result, but `write` requires a `Schema` and I only have the row type so I have to manually make a `Schema` type and type out the schema name again. It seems simpler and potentially less error prone to me to allow the user to use a row type when they have one. Then the call becomes `write(path, table, Schema(MyRowType))` (or just `MyRowType` with the `write` method change).

I guess a fairly direct OSS analogue is https://github.com/beacon-biosignals/Onda.jl/blob/master/src/signals.jl#L137, which requires rewriting the schema for `Signal`. When Onda bumps the `Signal` schema version, it'll have to be changed in multiple places, since it's written out explicitly rather than using the `Signal` type it already has.

jrevels: > MyRowType = @row(...), applying MyRowType to some rows, then wanting to write out the result, but write requires a Schema and I only have the row type so I have to manually make a Schema and type out the schema name again

I'm not sure I understand "and I only have the row type" - if you defined the row type, you should know the version of the schema you defined, no? In terms of changing it in multiple places - you'd have to do basically the same find-replace either way, right?

Does this come down to aesthetics for the "spelling of the invocation", or is there a different issue at play? The reason I'm inclined to think it's the former is because non-literal usages of this kind of path are a bit trickier in generic contexts for the aforementioned reason, so I'd assume it'd mostly get used "correctly" in the literal form, and in that case it essentially is an aesthetic choice

FWIW on the aesthetics front it makes more sense to me to steer table-level operations with `Schema` rather than row type constructors. Though I can see that `MyRowType` might read as less verbose than `Schema("my-schema@1")`

on the non-aesthetics front, the main two things i'd like to avoid/discourage for downstream users include:

- writing faux-generic code where behavior/correctness is predicated on whether or not a given row is a `Legolas.Row` vs. any compliant Tables.jl row 
- problematic assumptions/constraints on table types/parameters (e.g. a valid `Vector{NamedTuple{...}}` that can't be consumed properly because of assumptions about row type)

> I guess a fairly direct OSS analogue is https://github.com/beacon-biosignals/Onda.jl/blob/master/src/signals.jl#L137, which requires rewriting the schema for Signal. When Onda bumps the Signal schema version, it'll have to be changed in multiple places, since it's written out explicitly rather than using the Signal type it already has.

my preference is for these kinds of interfaces to be functional/value driven at the table level and not `Type{...}`-driven/row-driven for the above reasons

ararslan: > I'm not sure I understand "and I only have the row type" - if you defined the row type, you should know the version of the schema you defined, no?

Yes, you have the schema string, just not the `Schema` object you need in order to `write`. Constructing that requires rewriting the schema string instead of being able to query the row type for its schema (which would be particularly handy when you aren't the one who wrote the definition).

> In terms of changing it in multiple places - you'd have to do basically the same find-replace either way, right?

Perhaps I'm not understanding what you mean but with what I'm suggesting, you only ever write out the schema string once, so there's no find and replace to do other than modifying the version number at the definition site. So for example, Onda defines `Signal` and never has to write out `Schema("signal@1")` (or whatever its schema is) again; all further uses of `Signal` that need to know its schema can call `Schema(Signal)` or potentially use `Signal` itself directly.

> writing faux-generic code where behavior/correctness is predicated on whether or not a given row is a `Legolas.Row` vs. any compliant Tables.jl row

In this case, it's a strict addition of flexibility, not a change in behavior or correctness based on the input. The assumption is only that `Schema` can be called on whatever it is that was passed as that argument to `write`; users can still pass a `Schema` as before and it will behave the same, they just get the added convenience of being able to pass in things that have known schemas.

FWIW though I care less about the `write` method change than I do about being able to call `Schema` on rows/row types because the latter permits e.g. `write(path, table, Schema(MyRow))` which is good enough for me.

jrevels: > rewriting the schema string instead of being able to query the row type for its schema

This is what I think I mean by "this seems like a spelling/aesthetics thing": It seems like the difference is "you have to write `Schema("my-schema@1")`" vs. "you have to write `Schema(MyRowType)`". From a pure spelling perspective, I see no benefit to the latter and prefer the former

From a NOT pure spelling benefit (i.e. in generic code), I'm not sure I see any benefit but the downsides/increased surface for mistakes in generic code are pretty easy to imagine. 

Actually, there might be a better way to explain my reasoning/reluctance here: Let's assume we added this functionality, and walk through some hypothetical code review scenarios where it is used "in the wild."

- Scenario 1: I'm reviewing some code that invoked `Schema(T)` where `T` is a type parameter passed into the invocation context by an upstream caller. My review would be: "Let's refactor this to get rid of this invocation, and replace it with something else. Either `T` is a derived type parameter (e.g. an `eltype`) in which case this code is depending on things it shouldn't because its behavior will depend on whether an upstream type was inferred tightly enough (ref YASGuide's guidance on type parameter dependence), or it is a concrete type, in which case it's non-generic to other valid Tables.jl row types in a manner that's opaque to the caller"

- Scenario 2: I'm reviewing some code that invokes `Schema(row)` where `row` is a row value (in the Tables.jl sense) passed into the invocation context by an upstream caller. My review would be: "Let's refactor this to get rid of this invocation, and replace it with something else; this code now assumes that `row` is a `Legolas.Row`, which is an unnecessary constraint that is also likely to make usages of this code fragile/annoying in most contexts
 
 - Scenario 3: I'm reviewing some code that invokes `Schema(MyRowType)` where `MyRowType` is a literal type name/alias. My review would be: "Let's spell this as `Schema("my-schema@1")` instead, since it reads better and also surfaces the schema version at the callsite. Or, if we instead want the schema version to be driven by some upstream API function (that itself is hopefully SemVer'd), we can call that function instead."

Unless I'm missing something, these scenarios basically cover all usage cases for this functionality (please correct if I'm wrong!) and in every scenario my suggestion would be to refactor away the usage of this functionality; thus, my conclusion is that it means we probably shouldn't add this functionality

Does that make sense? or I might still be missing something interesting going on here

jrevels: >  there's more room for error if you need to remember and parrot the exact schema string and version rather than using an existing variable that stores it.

This is what I mean by it being a "spelling issue" - you're either "parroting" `"my-schema@1"` or "parroting" `MyRowType`. The typo mistake could happen just as easily either way.  The only difference in our scenario 3 is that a `MyMisspelledRowType` will error slightly earlier than `"my-schema@1"`. If scenarios 1/2 were valid, I'd be closer to agreeing with you, because in those scenarios the earlier errors could be "early enough" to actually be useful; but only looking at scenario 3, there's not a real benefit AFAICT. And actually, there's another class of errors to consider: that whatever literal is used is spelled correctly at the callsite, but that the prerequisite code hasn't been loaded. In that case, `"my-schema@1"` will result in a WAY more user-friendly error (since this case is so common, we added a special error path for it https://github.com/beacon-biosignals/Legolas.jl/pull/28).

Finally, even if we didn't align on the "spelling perspective" presented above, if you're seeking to use a constant binding to identify a schema rather than a literal string, then I'd say just do that instead of punning on a row type.

```
const MY_SCHEMA = Legolas.Schema("my-schema@1")
```

Though as mentioned previously, if we were to do this and thus remove callsite-specification of the schema version (not usually my preference), it's probably better to expose a functional interface to the hardcoded Schema (e.g. `do_with_my_schema(...)` or even just `my_schema()`) rather than a constant binding to it.

IIUC it seems like any of the above would give you any of the benefits you'd get with `Schema(::Type{<:Row})`, but without the downsides

jrevels:
I think the "preferred approach" to solving the problem that motivates this PR is captured in https://github.com/beacon-biosignals/Legolas.jl/pull/29#discussion_r772377881, so I'm going to go ahead and close. Can continue discussion in an issue if that approach has doesn't cover some aspect of the problem

PR #33, opened 2022-01-28
Update to codecov-action v2

kimlaberinto:
As mentioned in: https://github.com/codecov/codecov-action , codecov-action@v1 will be deprecated/sunset on February 1, 2022.

There are a few breaking changes mentioned in the codecov-action README.md, but I don't think this workflow uses any of the breaking changes mentioned.


jrevels:
Thanks Kim!

kimlaberinto:
You're welcome :D

Issue #34, opened 2022-02-01
support pluggable/arbitrary (de)serialization formats/targets (CSV, JSON, YAML, etc.)


jrevels:
From the [FAQ](https://beacon-biosignals.github.io/Legolas.jl/stable/faq/):

> Why does Legolas.jl support Arrow as a (de)serialization target, but not, say, JSON?
>
> Technically, Legolas.jl's core Row/Schema functionality is totally agnostic to (de)serialization and could be useful for anybody who wants to generate new Tables.AbstractRow types.
>
> Otherwise, with regards to (de)serialization-specific functionality, Beacon has put effort into ensuring Legolas.jl works well with Arrow.jl "by default" simply because we're heavy users of the Arrow format. There's nothing stopping users from composing the package with JSON3.jl or other packages.

Legolas rows are valid Tables.jl AbstractRows, so can already be fed into any package that (de)serializes Tables.jl tables.

The only "Legolas-specific" bit one might want to be captured is the schema metadata (e.g. the fully qualified schema string), in order to enable consumers to load tables into the appropriate schema. Arrow has a convenient custom table-level metadata mechanism we use for this currently; note that regardless of what we support as a result of this issue, we could still specialize on Arrow and continue to use its custom table-level metadata instead of some worse generic fallback.

You can imagine defining a little hook in Legolas' API that enables authors to define how to (de)serialize schema metadata to any given file format.

However, it'd be nice to not ever really need to build up a library of "supported formats" and instead have a generic approach. There are three ways I can imagine doing this that are fully agnostic, and one additional way that is fully agnostic to the format but still specialized on the storage system:

1. Simply write the Legolas metadata as a newline-terminated magic string/header before the table content, regardless of format. This is a little bit annoying for Legolas-unaware consumers of Legolas-written data (e.g. an arbitrary JSON reader might break), but at least it'd be pretty discoverable. 

2. Always encode the Legolas metadata as the first row in the table; e.g. `(a=[1,2,3], b=[5,6,7])` would be written out like `(a=["schema@1",1,2,3], b=[nothing,5,6,7])`. This solves the portability problem of the previous approach, but would mean that we couldn't guarantee homogenous column eltypes for the whole table for Legolas-unaware consumers that don't know to "skip" the first row

3. Just add a whole column to the table that contains the schema string duplicated for each row. This solves the previous two problems, but is a bit wasteful. but maybe not too bad given that the bloat is constant w.r.t. row size, and is easily compressible.

[...]

PR #38, opened 2022-04-20
tweak UnknownSchemaError language

ericphanson:
In some Beacon-internal code, I'm using TOML as a serialization format for small tables. You can get UnknownSchemaErrors, which are then confusing if they tell you to `Arrow.Table` it.

Maybe this is one tiny step towards #34

### DIFF
--- a/src/rows.jl
+++ b/src/rows.jl
@@ -119,7 +119,7 @@ function Base.showerror(io::IO, e::UnknownSchemaError)
               your environment is as expected).
 
               Note that if you're in this particular situation, you can still load
-              the raw Arrow table as-is via `Arrow.Table`.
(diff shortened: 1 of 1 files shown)


src/rows.jl
```diff
@@ -119,7 +119,7 @@ function Base.showerror(io::IO, e::UnknownSchemaError)
               your environment is as expected).
 
               Note that if you're in this particular situation, you can still load
-              the raw Arrow table as-is via `Arrow.Table`.
+              the raw table as-is. For example, to load an Arrow table, call `Arrow.Table(path)`.
```
jrevels, line 122:
```suggestion
              the raw table as-is without Legolas; e.g., to load an Arrow table, call `Arrow.Table(path)`.
```

slightly clearer/more helpful for users who might not understand what `as-is` means

PR #45, opened 2022-05-17
Add `construct` function

omus:
Method ensure compatibility with types which do not define constructing instances of themselves. For example with `Time`:

```julia
julia> Time(Time(0))
ERROR: MethodError: no method matching Int64(::Time)
```

We can work around this with `Legolas.lift(x -> x isa Time ? x : Time(x), y)` but with this PR we can simply do `Legolas.lift_type(Time, y)`.

### DIFF
--- a/src/lift.jl
+++ b/src/lift.jl
@@ -21,19 +21,23 @@ lift(f) = Base.Fix1(lift, f)
 
 
 """
-    lift_type(T::Type, x)
+    construct(T::Type, x)
 
-Construct `T(x)` unless `x` is of type `T` or `Missing` then `x` itself will be returned.
-If `x` is of type `Nothing` then the value `missing` will be returned instead.
+Construct `T(x)` unless `x` is of type `T`, in which case return `x` itself. Useful in
+conjunction with the [`lift`](@ref) function for types which don't have a constructor which
+accepts instances of itself (e.g. `T(::T)`).
 
-This is particularly useful when handling values from `Arrow.Table`, whose null values
-may present as either `missing` or `nothing` depending on how the table itself was
-originally constructed.
+## Examples
+```jldoctest
+julia> using Legolas: lift, construct
 
-See also: [`lift`](@ref)
+julia> lift(Some, Some(1))
+Some(Some(1))
+
+julia> lift(construct(Some), Some(1))
+Some(1)
+```
 """
-lift_type(T::Type, x) = T(x)
-lift_type(::Type{T}, x::T) where T = x
-lift_type(::Type, ::Union{Nothing,Missing}) = missing
-
-lift_type(T::Type) = Base.Fix1(lift_type, T)
+construct(T::Type, x) = T(x)
+construct(::Type{T}, x::T) where T = x
+construct(T::Type) = Base.Fix1(construct, T)
--- a/src/lift.jl
+++ b/src/lift.jl
@@ -55,5 +55,5 @@ Some(1)
 ```
 """
 construct(T::Type, x) = T(x)
-construct(::Type{T}, x::T) where T = x
+construct(::Type{T}, x::T) where {T} = x
 construct(T::Type) = Base.Fix1(construct, T)
--- a/src/lift.jl
+++ b/src/lift.jl
@@ -9,8 +9,8 @@ originally constructed.
 
 See also: [`lift_type`](@ref)
 """
-lift(::Any, ::Union{Nothing,Missing}) = missing
 lift(f, x) = f(x)
+lift(::Any, ::Union{Nothing,Missing}) = missing
 
 """
     lift(f)
--- a/src/lift.jl
+++ b/src/lift.jl
@@ -7,7 +7,7 @@ This is particularly useful when handling values from `Arrow.Table`, whose null 
 may present as either `missing` or `nothing` depending on how the table itself was
 originally constructed.
 
-See also: [`lift_type`](@ref)
+See also: [`construct`](@ref)
 """
 lift(f, x) = f(x)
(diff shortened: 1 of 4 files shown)


ericphanson:
Is this ambiguous for `Legloas.lift(Time, missing)`? Might be worth a test just in case

jrevels:
I'm semi-in-favor of this just because it can get so annoying otherwise, but it definitely is a breaking change. and I think as a user, my naive assumption would be that `lift` would not have any automagical special-casing for specific `f` types, and it might bite me if it does

we can instead avoid the breaking change and automagic behavior by defining a function like

```jl
if_needed(T) = x -> x isa T ? x : T(x)
```

(horrible name, please pick a better one lol - would be good if it was short but still clear)

...which could then be used like: 

```jl
lift(if_needed(T), x)
```

introducing this would be a non-breaking change and nicer to keep `lift` very "dumb"

ararslan:
I think in an ideal world, i.e. one without https://github.com/apache/arrow-julia/issues/258, we would be able to deprecate `lift` in favor of `passmissing` from Missings. Adding additional methods to `lift` with specific behavior would make that transition a bit less seamless (though the handling of `nothing` does that already). Unfortunately from an ergonomics perspective though, `passmissing` doesn't fare any better in terms of handling type constructors without identity methods.

Though similar to what Jarrett said, I think that as a user, more than anything I would expect that in `lift(f, x)`, `f(x)` would actually be called if `x !== missing`. The documentation does say that `f(x)` is returned so if we go forward with the change as written then we should update the the docstring accordingly.

> The implementation of this probably would result in a new function name being cleaner.

`lift_type`? Just `lift` but for type constructors.
```julia
lift_type(::Type{T}, x::T) where {T} = x
lift_type(::Type{T}, x) where {T} = T(x)
lift_type(::Type, ::Union{Missing,Nothing}) = missing
```

omus:
> `lift_type`? Just `lift` but for type constructors.

I thought of that same name as well as the more cheeky `uplift` (less descriptive). As it turns out Jarrett's `if_needed` is just the unary version of `lift_type`:

```julia
lift(lift_type(T), x)
```
so just using `lift_type` directly seems more sensible

omus:
> i'm still a fan of the higher order function that can just be composed with existing `lift`

For completeness here's what that could look like:
```julia
julia> using Dates, Legolas

julia> construct(T::Type, x) = T(x)
construct (generic function with 1 method)

julia> construct(::Type{T}, x::T) where T = x
construct (generic function with 2 methods)

julia> construct(T::Type) = Base.Fix1(construct, T)
construct (generic function with 3 methods)

julia> Legolas.lift(construct(Time), Time(0))
00:00:00
```

src/lift.jl
```diff
@@ -0,0 +1,59 @@
...
+julia> lift(Some, Some(1))
+Some(Some(1))
+
+julia> lift(construct(Some), Some(1))
+Some(1)
+```
+"""
+construct(T::Type, x) = T(x)
+construct(::Type{T}, x::T) where T = x
+construct(T::Type) = Base.Fix1(construct, T)
```
jrevels, line 59:
```suggestion
construct(T, x) = T(x)
construct(::Type{T}, x::T) where {T} = x
construct(T) = Base.Fix1(construct, T)
```

IMO there's no need for the `::Type` constraint on `T` for the fallback case

(also `where T` -> `where {T}` for YASGuide purposes)

omus: Although the type constraint isn't needed for functionality this does generalize the function such that the docstring is no accurate (`T` may not be a type) and the not using a type for `T` would also result in me no longer wanting to use the name `construct` for the function as construction/constructors are tied to types

jrevels: > generalize the function such that the docstring is no accurate (T may not be a type) 

yeah, my suggestion here IS to generalize the function such that `T` may not be a type (if you wanted could rename to `f` too)

> construction/constructors are tied to types

i see no reason to restrict it to types vs. any generic non-type callable

but i'm also fine moving forward with this PR without generalizing if there's some concern there

(aside: this is such a bikesheddable bite-sized PR @omus , sorry 😂)

omus: > this is such a bikesheddable bite-sized PR @omus , sorry 😂

Very true. I think we can just proceed with this PR as it currently exists. If we want to generalize this function further I think the name should change which means we can easily replace this function with something else in the future.

src/lift.jl
```diff
@@ -0,0 +1,59 @@
+"""
+    lift(f, x)
+
+Return `f(x)` unless `x isa Union{Nothing,Missing}`, in which case return `missing`.
+
+This is particularly useful when handling values from `Arrow.Table`, whose null values
+may present as either `missing` or `nothing` depending on how the table itself was
+originally constructed.
+
+See also: [`lift_type`](@ref)
```
ararslan, line 10:
```suggestion
See also: [`construct`](@ref)
```

src/lift.jl
```diff
@@ -0,0 +1,59 @@
...
+julia> using Legolas: lift, construct
+
+julia> lift(Some, Some(1))
+Some(Some(1))
+
+julia> lift(construct(Some), Some(1))
+Some(1)
+```
+"""
+construct(T::Type, x) = T(x)
```
ararslan, line 57: Just for my own edification, not suggesting a change: Why `T::Type` rather than `::Type{T}`? To avoid forcing specialization on `T`?

jrevels (approved):

src/lift.jl
omus: > To avoid forcing specialization on `T`?

Exactly

Issue #41, opened 2022-05-04
Add function that returns a Schema's cols


hannahilea:
e.g., `fieldnames` or or `propertynames` or `getproperties`, etc

ararslan:
The way things are set up currently, a Legolas `Schema` object has no knowledge at all of the actual schema it describes; all it contains is a name and version number. (In contrast, a Tables `Schema` describes the actual schema but has no notion of versioning.) You actually need a Legolas `Row` in order to determine the fields, as that stores both the schema and the fields, but that may also contain other fields not required per the `Row`'s initial `@row` definition.

Well, there is... one way...
```julia
julia> function expected_names(::T) where {T<:Schema}
           for method in methods(Legolas._transform)
               params = (method.sig.parameters...,)
               length(params) >= 2 && params[2] <: T || continue
               names = Base.kwarg_decl(method)
               filter!(x -> !endswith(String(x), "..."), names)
               return names
           end
           throw(ArgumentError("can't determine expected names for $T"))
       end
expected_names (generic function with 1 method)

julia> @row("cat@1", name::String, breed::String, age::Int);

julia> expected_names(Schema("cat", 1))
3-element Vector{Symbol}:
 :name
 :breed
 :age
```
The only place the expected fields get exposed outside of a method body is in the keyword arguments for the internal function `_transform`, so if you want to know the names, you can look for the matching method of that function. :trollface:

ararslan:
More seriously, something like
```diff
diff --git a/src/rows.jl b/src/rows.jl
index 92190b3..1cdbc91 100644
--- a/src/rows.jl
+++ b/src/rows.jl
@@ -181,6 +181,10 @@ metadata for table written via [`Legolas.write`](@ref).
 """
 schema_qualified_string(s::Legolas.Schema) = throw(UnknownSchemaError(s))

+function schema_field_names end
+
+function schema_field_types end
+
 #####
 ##### Row
 #####
@@ -315,6 +319,13 @@ macro row(schema_expr, fields...)
             return _validate(tables_schema, legolas_schema)
         end

+        function Legolas.schema_field_names(::$schema_type)
+            return $(Expr(:vect, map(f -> QuoteNode(f.args[1].args[1]), fields)...))
+        end
+
+        function Legolas.schema_field_types(::$schema_type)
+            return $(Expr(:vect, map(f -> esc(f.args[1].args[2]), fields)...))
+        end

[...]

PR #52, opened 2022-08-09
Add column name/utility accessors for `Legolas.Schema`

OTDE:
This PR addresses #41. It adds two functions:
```
    schema_field_names(::Type{<:Legolas.Schema})

Get a tuple with the names of the fields of this `Legolas.Schema`, including names that
have been inherited from this `Legolas.Schema`'s parent schema.
```
```
    schema_field_types(::Legolas.Schema{name,version})

Get a tuple with the types of the fields of this `Legolas.Schema`, including types of fields that
have been inherited from this `Legolas.Schema`'s parent schema.
```

Tests have been added to make sure these functions behave like we expect.

```julia

Parent = @row("parent@1",
              first_parent_field::Int=1,
              second_parent_field::String="second")

# returns (:first_parent_field, :second_parent_field)
Legolas.schema_field_names(Legolas.Schema{:parent,1}) 
Legolas.schema_field_names(Legolas.Schema("parent@1"))
Legolas.schema_field_names(Parent())
Legolas.schema_field_names(Parent)

# returns (Int, String)
Legolas.schema_field_types(Schema{:parent,1})
Legolas.schema_field_types(Schema("parent@1"))
Legolas.schema_field_types(Parent())
Legolas.schema_field_types(Parent)
```

This includes tests for schema inheritance as well:
```julia
Child = @row("child@1" > "parent@1",
              first_child_field::Symbol=:first,
              second_child_field="I can be anything")

# returns (:first_child_field, :second_child_field, :first_parent_field, :second_parent_field)
Legolas.schema_field_names(Child) 

# returns (Symbol, Any, Int, String)
Legolas.schema_field_types(Child)
```

### DIFF
--- a/src/rows.jl
+++ b/src/rows.jl
@@ -105,25 +105,22 @@ struct UnknownSchemaError <: Exception
 end
 
 function Base.showerror(io::IO, e::UnknownSchemaError)
-    print(
-        io,
-        """
-        encountered unknown `Legolas.Schema` type: $(e.schema)
+    print(io, """
+              encountered unknown `Legolas.Schema` type: $(e.schema)
 
-        This generally indicates that this schema has not been defined (i.e.
-        the schema's corresponding `@row` statement has not been executed) in
-        the current Julia session.
+              This generally indicates that this schema has not been defined (i.e.
+              the schema's corresponding `@row` statement has not been executed) in
+              the current Julia session.
 
-        In practice, this can arise if you try to read a Legolas table with a
-        prescribed schema, but haven't actually loaded the schema definition
-        (or commonly, haven't loaded the dependency that contains the schema
-        definition - check the versions of loaded packages/modules to confirm
-        your environment is as expected).
+              In practice, this can arise if you try to read a Legolas table with a
+              prescribed schema, but haven't actually loaded the schema definition
+              (or commonly, haven't loaded the dependency that contains the schema
+              definition - check the versions of loaded packages/modules to confirm
+              your environment is as expected).
 
-        Note that if you're in this particular situation, you can still load
-        the raw table as-is without Legolas; e.g., to load an Arrow table, call `Arrow.Table(path)`.
-        """
-    )
+              Note that if you're in this particular situation, you can still load
+              the raw table as-is without Legolas; e.g., to load an Arrow table, call `Arrow.Table(path)`.
+              """)
     return nothing
 end
 
--- a/src/rows.jl
+++ b/src/rows.jl
@@ -239,6 +239,7 @@ have been inherited from this `Legolas.Schema`'s parent schema.
 schema_field_names(::Type{S}) where {S<:Legolas.Schema} = throw(UnknownSchemaError(S()))
 schema_field_names(s::Legolas.Schema) = schema_field_names(typeof(s))
 schema_field_names(::Legolas.Row{S}) where {S} = schema_field_names(S)
+schema_field_names(::Type{<:Legolas.Row{S}}) where {S} = schema_field_names(S)
 
 """
     schema_field_types(::Legolas.Schema{name,version})
@@ -249,6 +250,7 @@ have been inherited from this `Legolas.Schema`'s parent schema.
 schema_field_types(::Type{S}) where {S<:Legolas.Schema} = throw(UnknownSchemaError(S()))
 schema_field_types(s::Legolas.Schema) = schema_field_types(typeof(s))
 schema_field_types(::Legolas.Row{S}) where {S} = schema_field_types(S)
+schema_field_types(::Type{<:Legolas.Row{S}}) where {S} = schema_field_types(S)
 
 function _parse_schema_expr(x)
     if x isa Expr && x.head == :call && x.args[1] == :> && length(x.args) == 3
--- a/src/rows.jl
+++ b/src/rows.jl
@@ -105,22 +105,25 @@ struct UnknownSchemaError <: Exception
 end
 
 function Base.showerror(io::IO, e::UnknownSchemaError)
-    print(io, """
-              encountered unknown `Legolas.Schema` type: $(e.schema)
+    print(
+        io,
+        """
+        encountered unknown `Legolas.Schema` type: $(e.schema)
 
-              This generally indicates that this schema has not been defined (i.e.
-              the schema's corresponding `@row` statement has not been executed) in
-              the current Julia session.
+        This generally indicates that this schema has not been defined (i.e.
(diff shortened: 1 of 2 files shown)


src/rows.jl
```diff
@@ -227,6 +230,28 @@ function Base.show(io::IO, row::Row)
...
+"""
+    schema_field_names(::Type{<:Legolas.Schema})
+
+Get a tuple with the names of the fields of this `Legolas.Schema`, including names that
+have been inherited from this `Legolas.Schema`'s parent schema.
+"""
+schema_field_names(::Type{S}) where {S<:Legolas.Schema} = throw(UnknownSchemaError(S()))
+schema_field_names(s::Legolas.Schema) = schema_field_names(typeof(s))
+schema_field_names(::Legolas.Row{S}) where {S} = schema_field_names(S)
+schema_field_names(::Type{<:Legolas.Row{S}}) where {S} = schema_field_names(S)
```
ararslan, line 239:
```suggestion
schema_field_names(::Type{S}) where {S<:Schema} = throw(UnknownSchemaError(S()))
schema_field_names(s::Schema) = schema_field_names(typeof(s))
schema_field_names(::Row{S}) where {S} = schema_field_names(S)
schema_field_names(::Type{<:Row{S}}) where {S} = schema_field_names(S)
```
There's no need to qualify things with the name of the module you're currently in (except in very rare circumstances).

src/rows.jl
```diff
@@ -227,6 +227,28 @@ function Base.show(io::IO, row::Row)
...
+"""
+    schema_field_types(::Legolas.Schema{name,version})
+
+Get a tuple with the types of the fields of this `Legolas.Schema`, including types of fields that
+have been inherited from this `Legolas.Schema`'s parent schema.
+"""
+schema_field_types(::Type{S}) where {S<:Legolas.Schema} = throw(UnknownSchemaError(S()))
+schema_field_types(s::Legolas.Schema) = schema_field_types(typeof(s))
+schema_field_types(::Legolas.Row{S}) where {S} = schema_field_types(S)
+schema_field_types(::Type{<:Legolas.Row{S}}) where {S} = schema_field_types(S)
```
ararslan, line 250:
```suggestion
schema_field_types(::Type{S}) where {S<:Schema} = throw(UnknownSchemaError(S()))
schema_field_types(s::Schema) = schema_field_types(typeof(s))
schema_field_types(::Row{S}) where {S} = schema_field_types(S)
schema_field_types(::Type{<:Row{S}}) where {S} = schema_field_types(S)
```
(same as previous comment)

test/runtests.jl
```diff
@@ -176,15 +176,54 @@ end
...
+    @test Legolas.schema_field_names(Parent) == parent_fields
+
+    @test Legolas.schema_field_types(Schema{:parent,1}) == parent_field_types
+    @test Legolas.schema_field_types(Schema("parent@1")) == parent_field_types
+    @test Legolas.schema_field_types(Parent()) == parent_field_types
+    @test Legolas.schema_field_types(Parent) == parent_field_types
+
+    Child = @row("child@1" > "parent@1",
+                  first_child_field::Symbol=:first,
+                  second_child_field="I can be anything")
```
ararslan, line 199:
```suggestion
                 first_child_field::Symbol=:first,
                 second_child_field="I can be anything")
```

src/rows.jl
```diff
@@ -227,6 +227,28 @@ function Base.show(io::IO, row::Row)
     return nothing
 end
 
+"""
+    schema_field_names(::Type{<:Legolas.Schema})
+
+Get a tuple with the names of the fields of this `Legolas.Schema`, including names that
+have been inherited from this `Legolas.Schema`'s parent schema.
+"""
+schema_field_names(::Type{S}) where {S<:Schema} = throw(UnknownSchemaError(S()))
```
ararslan, line 236: I actually think the `MethodError` you'd get from removing this fallback method is more correct than an `UnknownSchemaError`; it could be that a schema exists (and thus is "known") but a method for `schema_field_names` simply wasn't defined for it.

src/rows.jl
```diff
@@ -105,22 +105,25 @@ struct UnknownSchemaError <: Exception
...
+        In practice, this can arise if you try to read a Legolas table with a
+        prescribed schema, but haven't actually loaded the schema definition
+        (or commonly, haven't loaded the dependency that contains the schema
+        definition - check the versions of loaded packages/modules to confirm
+        your environment is as expected).
+
+        Note that if you're in this particular situation, you can still load
+        the raw table as-is without Legolas; e.g., to load an Arrow table, call `Arrow.Table(path)`.
+        """
+    )
```
jrevels, line 126: fix style?

src/rows.jl
jrevels: > it could be that a schema exists (and thus is "known") but a method for schema_field_names simply wasn't defined for it.

Hmmm how so?

I maybe skimmed too fast but using `UnknownSchemaError` here seems to be consistent with the other usages of it (and is useful here for the same reason it's useful in those other contexts)

src/rows.jl
```diff
@@ -227,6 +230,26 @@ function Base.show(io::IO, row::Row)
...
+
+"""
+    schema_field_types(::Legolas.Schema{name,version})
+
+Get a tuple with the types of the fields of this `Legolas.Schema`, including types of fields that
+have been inherited from this `Legolas.Schema`'s parent schema.
+"""
+schema_field_types(s::Schema) = schema_field_types(typeof(s))
+schema_field_types(::Row{S}) where {S} = schema_field_types(S)
+schema_field_types(::Type{<:Row{S}}) where {S} = schema_field_types(S)
```
jrevels, line 251: Instead of `schema_field_names` / `schema_field_types`, should we define `Tables.Schema(::Legolas.Schema)` or something similar?

https://tables.juliadata.org/stable/#Tables.Schema is the natural object to use for this purpose in Tables.jl ecosystem IIUC

note that I think `Tables.schema(::Vector{<:Legolas.Row})`might already "just work" on `main` (the lowercase `schema` is a generic accessor). Would be worth checking.

Also we should keep in mind that Legolas Schema fields are explicitly unordered - might be best to call that out in the function docs as well

jrevels (changes requested):
Largely in favor!

see above comments for change suggestions

also before merging we should bump the package version as part of the PR

src/rows.jl
ararslan: > Hmmm how so?

Very hypothetical case of defining a schema without using `@row`.

If it's consistent and useful then that's good enough for me.

jrevels:
superseded by #54

PR #62, opened 2022-11-03
add missing docstring for Legolas.accepted_field_type

jrevels:
This also tightens the fallback type constraint on `T` from `Any` to `Type`. I don't view this a breaking change since this function would make no sense with non-type arguments anyway (if being passed non-type arguments, there's probably a silent bug happening elsewhere anyway...), but if controversial, i can just remove that part


ericphanson (approved):
lgtm, thanks!

Issue #57, opened 2022-10-27
add documentation to aid upgrading from v0.4 to v0.5


jrevels:
ref https://github.com/beacon-biosignals/Legolas.jl/pull/54#issuecomment-1293811402

PR #71, opened 2022-11-22
start upgrade guide, and add pieces from LegolasFlux upgrade

ericphanson:
when @kleinschmidt and I upgraded LegolasFlux to Legolas v0.5, we encountered [several points](https://github.com/beacon-biosignals/LegolasFlux.jl/pull/21#pullrequestreview-1188712305) that we thought would be useful to as upgrade guidance for our users. All of those points seem general and not specific to LegolasFlux however, so I think it would be more appropriate to put it in Legolas's documentation itself.

This adds incomplete documentation (since I don't think this is a full upgrade guide), but it is deliberately called out as such. IMO it would be better to incrementally add pieces as folks find things rather than block such guidance on being complete.

ref https://github.com/beacon-biosignals/Legolas.jl/issues/57


docs/src/upgrade.md
```diff
@@ -0,0 +1,16 @@
...
+* Constructors for structs generated by `@version` calls discard "extra" columns, whereas previously under Legolas v0.4, `@row`-generated constructors kept any "extra" columns. It may be advisable to add previously used "extra" columns into the schema itself (probably declared as `::Union{Missing,T}` to allow them to be missing).
+
+
+## Deserializing old tables with Legolas v0.5
+
+Generally, tables serialized with Legolas v0.4 can be de-serialized with Legolas v0.5, making it only a "code-breaking" change, rather than a "data-breaking" change. However, it is strongly suggested to have reference tests with checked in (pre-Legolas v0.5) serialized tables which are deserialized and verified during the tests, in order to be sure.
```
ericphanson, line 14:
```suggestion
Generally, tables serialized with earlier versions of Legolas can be de-serialized with Legolas v0.5, making it only a "code-breaking" change, rather than a "data-breaking" change. However, it is strongly suggested to have reference tests with checked in (pre-Legolas v0.5) serialized tables which are deserialized and verified during the tests, in order to be sure.
```

docs/src/upgrade.md
```diff
@@ -0,0 +1,16 @@
...
+
+
+## Deserializing old tables with Legolas v0.5
+
+Generally, tables serialized with earlier versions of Legolas can be de-serialized with Legolas v0.5, making it only a "code-breaking" change, rather than a "data-breaking" change. However, it is strongly suggested to have reference tests with checked in (pre-Legolas v0.5) serialized tables which are deserialized and verified during the tests, in order to be sure.
+
+Additionally, tables with nested Legolas schemas (i.e. a row has a column whose elements are themselves Legolas rows) will need special handling to deserialize under Legolas v0.5. Recall that under Legolas v0.4, row types can contain arbitrary "extra" columns not specified in the schema. Under Legolas v0.5, their analog, namely structs created by the `@version` macro, only allow columns specified in their schema. Therefore, one must decide what to do with any "extra" columns when deserializing older tables. These columns can generally be handled in the constructor, as in [this LegolasFlux example](https://github.com/beacon-biosignals/LegolasFlux.jl/blob/53c677848c6b65e5158ef2d43dd5f7eab174892e/examples/digits.jl#L64-L84).
```
kleinschmidt, line 16: this example isn't really about handling extra columns right?  they'll still be discarded by the `DigitsModelV1` constructor.  AFAICT there isn't really a way to BOTH have nested row-struct-typed-fields in a schema AND support extra columns that might be present in that field's nested row serialized under old versions of Legolas.

kleinschmidt (approved):
looks good with one little quibble!  I agree this is a good place for it to live

docs/src/upgrade.md
ericphanson: True, yeah. What I wanted to say was something like "you have to figure out what to do with them. There's a place you get to write code to do that, and it's in the constructor". And I guess hinting at you can do whatever you need to do there, e.g. add some weird `extras::NamedTuple` column to your schema and populate it this way, or throw an error and tell the user to manually do something, or warn or something, idk.

hannahilea:
:glue: ❤️ :teamwork-makes-the-dreamwork:

docs/src/upgrade.md
ericphanson:
```suggestion
Additionally, tables with nested Legolas schemas (i.e. a row has a column whose elements are themselves Legolas rows) will need special handling to deserialize under Legolas v0.5. Recall that under Legolas v0.4, row types can contain arbitrary "extra" columns not specified in the schema. Under Legolas v0.5, their analog, namely structs created by the `@version` macro, only allow columns specified in their schema. Therefore, one must decide what to do with any "extra" columns when deserializing older tables, and use the constructor to old Legolas rows to the new types. For example, [this LegolasFlux example](https://github.com/beacon-biosignals/LegolasFlux.jl/blob/53c677848c6b65e5158ef2d43dd5f7eab174892e/examples/digits.jl#L64-L84) uses a function `compat_config` to handle old Legolas rows, but does not add any handling for "extra" columns, which will be discarded if present. In that scenario, extra columns could be handled by throwing an error, a warning, or expanding the schema (e.g. with `extras::Union{Missing, NamedTuple}`).
```

docs/src/upgrade.md
```diff
@@ -0,0 +1,16 @@
+# Upgrading from Legolas v0.4 to v0.5
+
+This guide is incomplete; please add to it if you encounter items which would help other upgraders along their journey.
+
+## Some main changes to be aware of
+
+* In Legolas v0.4, type parameters were implicit in the row types; for example, a column `y::Real` in a `Legolas.@row` declaration would be represented by `y::T` with the constraint `T<:Real`, rather than using the abstract type `Real` as a type declaration for `y`. Now, all type parameters need to manually specified with the `y::(<:Real)` syntax in [`@version`](@ref).
```
jrevels, line 7:
```suggestion
* In Legolas v0.4, every `Legolas.Row` field's type was available as a type parameter of `Legolas.Row`; for example, the type of a field `y` specified as `y::Real` in a `Legolas.@row` declaration would be surfaced like `Legolas.Row{..., NamedTuple{(...,:y,...),Tuple{...,typeof(y),...}}`. In Legolas v0.5, the schema version author controls which fields have their types surfaced as type parameters in Legolas-generated record types via the `field::(<:F)` syntax in [`@version`](@ref).
```

edit to be a bit more concrete / clearer

docs/src/upgrade.md
```diff
@@ -0,0 +1,16 @@
...
+    * Additionally, to include type parameters associated to fields in a parent schema, they must be re-declared in the child schema. For example, the package LegolasFlux declares a `ModelV1` version with a field `weights::(<:Union{Missing,Weights})`. LegolasFlux includes an [example](https://github.com/beacon-biosignals/LegolasFlux.jl/blob/53c677848c6b65e5158ef2d43dd5f7eab174892e/examples/digits.jl#L78-L80) with a schema extension `DigitsRowV1` which extends `ModelV1`. This `@version` call must re-declare the field `weights` to be parametric in order for the `DigitsRowV1` struct to also have a type parameter for this field.
+* Constructors for structs generated by `@version` calls discard "extra" columns, whereas previously under Legolas v0.4, `@row`-generated constructors kept any "extra" columns. It may be advisable to add previously used "extra" columns into the schema itself (probably declared as `::Union{Missing,T}` to allow them to be missing).
```
jrevels, line 9:
```suggestion
* In Legolas v0.4,  `@row`-generated `Legolas.Row` constructors accepted and propagated any non-schema-required fields provided by the caller. In Legolas v0.5, `@version`-generated record type constructors will discard any non-schema-required fields provided by the caller. When upgrading code that formerly "implicitly extended" a given schema version by propagating non-required fields, it is advisable to instead explicitly declare a new extension of the schema version to capture the propagated fields as required fields; or, if it makes more sense for a given use case, one may instead define a new schema version that adds these propagated fields as required fields directly to the schema (likely declared as `::Union{Missing,T}` to allow them to be missing)
```

it's more ideal in most cases to convert "implicit extensions" to "explicit extensions"; and if not, adding new required fields necessitates a schema version bump

docs/src/upgrade.md
```diff
@@ -0,0 +1,16 @@
...
+
+Additionally, tables with nested Legolas schemas (i.e. a row has a column whose elements are themselves Legolas rows) will need special handling to deserialize under Legolas v0.5. Recall that under Legolas v0.4, row types can contain arbitrary "extra" columns not specified in the schema. Under Legolas v0.5, their analog, namely structs created by the `@version` macro, only allow columns specified in their schema. Therefore, one must decide what to do with any "extra" columns when deserializing older tables, and use the constructor to old Legolas rows to the new types. For example, [this LegolasFlux example](https://github.com/beacon-biosignals/LegolasFlux.jl/blob/53c677848c6b65e5158ef2d43dd5f7eab174892e/examples/digits.jl#L64-L84) uses a function `compat_config` to handle old Legolas rows, but does not add any handling for "extra" columns, which will be discarded if present. In that scenario, extra columns could be handled by throwing an error, a warning, or expanding the schema (e.g. with `extras::Union{Missing, NamedTuple}`).
```
jrevels, line 16:
```suggestion
Additionally, serialized Arrow tables containing nested Legolas-v0.4-defined `Legolas.Row` values (i.e. a table that contains a row that has a field that is, itself, a `Legolas.Row` value, or contains such values) require special handling to deserialize under Legolas v0.5, if you wish users to be able to deserialize them with `Legolas.read` using the Legolas-v0.5-ready version of your package. Note that these tables are still deserializable as plain Arrow tables regardless, so it may not be worthwhile to provide a bespoke deprecation/compatibility pathway in the  Legolas-v0.5-ready version package unless your use case merits it (i.e. the impact surface would be high for your package's users). 

If you would like to provide such a pathway, though: 

Recall that under Legolas v0.4, `@row`-generated `Legolas.Row` constructors may accept and propagate arbitrary non-schema-required fields, whereas Legolas v0.5's `@version`-generated record types may only contain schema-required fields. Therefore, one must decide what to do with any non-required fields present in serialized `Legolas.Row` values upon deserialization. A common approach is to implement a deprecation/compatibility pathway within the relevant surrounding `@version` declaration. For example, [this LegolasFlux example](https://github.com/beacon-biosignals/LegolasFlux.jl/blob/53c677848c6b65e5158ef2d43dd5f7eab174892e/examples/digits.jl#L64-L84) uses a function `compat_config` to handle old `Legolas.Row` values, but does not add any handling for non-required fields, which will be discarded if present. If one did not want non-required fields to be discarded, these fields could be handled by throwing an error or warning, or defining a schema version extension that captured them, or defining a new version of the relevant schema to capture them (e.g. adding a field like `extras::Union{Missing, NamedTuple}`).
```

jrevels (changes requested):
LGTM pending folding in suggestions!

docs/src/upgrade.md
```diff
@@ -0,0 +1,16 @@
+# Upgrading from Legolas v0.4 to v0.5
+
+This guide is incomplete; please add to it if you encounter items which would help other upgraders along their journey.
```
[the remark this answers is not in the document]
ericphanson: how about this?
```suggestion
This guide is incomplete; please add to it if you encounter items which would help other upgraders along their journey.

See [here](https://github.com/beacon-biosignals/Legolas.jl/pull/54) for an example of a comprehensive log of changes from Legolas v0.4 to Legolas v0.5.
```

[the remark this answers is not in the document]
jrevels: 🤔 but it's not an example of some other thing? it's the PR's changelog

jrevels (approved):
not sure i understand the "an example of a comprehensive log of changes" change (it's not an example of a changelog, it's just the changelog that accompanies the PR) but doesn't matter either way lol

docs/src/upgrade.md
```diff
@@ -0,0 +1,22 @@
...
+
+See [here](https://github.com/beacon-biosignals/Legolas.jl/pull/54) for an example of a comprehensive log of changes from Legolas v0.4 to Legolas v0.5.
```
ericphanson, line 5:
```suggestion
See [here](https://github.com/beacon-biosignals/Legolas.jl/pull/54) for a comprehensive log of changes from Legolas v0.4 to Legolas v0.5.
```

ericphanson:
> not sure i understand the "an example of a comprehensive log of changes" change (it's not an example of a changelog, it's just the changelog that accompanies the PR) but doesn't matter either way lol

Oh oops, somehow I thought that link pointed at https://github.com/beacon-biosignals/LegolasFlux.jl/pull/21 as an upgrade example, in which case it wouldn't be a comprehensive log of changes to this package. But I see the link points to #54, which makes more sense, and the wording made sense as-is.

jrevels:
just realized it might be mildly useful for me to provide slightly more context on my thought process for one of [the suggested changes we folded in](https://github.com/beacon-biosignals/Legolas.jl/pull/71/commits/0f68aacf3f303f57e1cbfb403fc10ed898e1bdce) (and to surface for lurkers who might've missed the nested suggestion in the first place)  

```diff
- Constructors for structs generated by `@version` calls discard "extra" columns, 
- whereas previously under Legolas v0.4, `@row`-generated constructors kept any "extra" 
- columns. It may be advisable to add previously used "extra" columns into the schema itself 
- (probably declared as `::Union{Missing,T}` to allow them to be missing).
+ In Legolas v0.4,  `@row`-generated `Legolas.Row` constructors accepted and propagated any 
+ non-schema-required fields provided by the caller. In Legolas v0.5, `@version`-generated record 
+ type constructors will discard any non-schema-required fields provided by the caller. When upgrading 
+ code that formerly "implicitly extended" a given schema version by propagating non-required fields, it 
+ is advisable to instead explicitly declare a new extension of the schema version to capture the propagated 
+ fields as required fields; or, if it makes more sense for a given use case, one may instead define a new 
+ schema version that adds these propagated fields as required fields directly to the schema (likely 
+ declared  as `::Union{Missing,T}` to allow them to be missing).
```

tidbits to draw attention to:

- subtle nitpick, but since legolas doesn't really do anything at the table-level yet (besides validating the values/types of a table's rows), i defensively try to always use "fields" instead of "columns" just to avoid confusion (though the latter isn't necessarily incorrect depending on your perspective)

- the former advice could be misunderstood to mean "add a new required field to an existing schema version" (though it might not have intended that) but doing so would be a breaking change which necessitates a schema version bump

- generally the preferred path in Legolas v0.5 is to translate implicit/dynamic extensions into explicit/static extensions wherever reasonable, though in some cases introducing that might entail a tradeoff between "thrust things into the Right Way on Legolas v0.5" and the amount of breakage you're willing to propagate that may or may not be worthwhile depending on the scenario

PR #72, opened 2022-12-02
Minor tour changes

omus:


examples/tour.jl
```diff
@@ -120,7 +120,7 @@ fields = (a=1.0, b="hi", c=π, d=[1, 2, 3])
 #
 # - ...contain the associated schema version's required fields in any order
 # - ...elide required fields, in which case the constructor will assume them to be `missing`
-# - ...contain any other fields in addition to the required fields; such additional fields are simply ignored
+# - ...contain any other fields in addition to the required fields; such additional fields are simply dropped
```
jrevels, line 123:
```suggestion
# - ...contain any other fields in addition to the required fields; such additional fields are simply ignored 
#   by the constructor and are not propagated through to the resulting record.
```

figured it's better to just be a bit verbose here; "dropped" makes it sounds like we're removing fields post-hoc from a value that had those fields, as opposed to just ignoring them and not propagating them in the first place

jrevels (approved):

Issue #10, opened 2021-06-25
Overridable docstrings for `@version`-generated record types


ericphanson:
I think this could be done by generating a method for [Base.getdoc](https://docs.julialang.org/en/v1/manual/documentation/#Dynamic-documentation) in the `@row` macro.

omus:
I split out the autogenerated portion of the original issue into #75

PR #74, opened 2022-12-07
Support attaching docstrings to schema versions

omus:
Adds support for attaching docstrings to schema versions. I also attempted to also add support to `@schema` but it wasn't clear what construct the docstring should be attached to.

Fixes: #10

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -230,6 +230,17 @@ end
     x::Int
 end
 
+@schema "test.documented" Documented
+
+"""
+    DocumentedV1
+
+Very detailed documentation.
+"""
+@version DocumentedV1 begin
+    x
+end
+
 @testset "`Legolas.@version` and associated utilities for declared `Legolas.SchemaVersion`s" begin
     @testset "Legolas.SchemaVersionDeclarationError" begin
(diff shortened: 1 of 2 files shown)


test/runtests.jl
```diff
@@ -368,6 +379,11 @@ end
     roundtripped = Legolas.read(Legolas.tobuffer(tbl, NestedAgainV1SchemaVersion()))
     @test roundtripped.n[1] == NestedV1(; gc=GrandchildV1(r0_roundtripped), k="test")
     @test roundtripped.h[1] == 3
+
+    @testset "docstring support" begin
+        ds = Docs.docstr(Docs.Binding(@__MODULE__, :DocumentedV1))
```
christopher-dG, line 384: You could also get the docstring in a slightly less scary way with `string(@doc DocumentedV1)`

ericphanson:
fixes https://github.com/beacon-biosignals/Legolas.jl/issues/10?

test/runtests.jl
omus: Thanks!

omus:
> fixes #10?

Partially addresses #10. We're not doing autogenerated docstrings here. We may want to promote a more human readable form of docstring like I am suggesting in: https://github.com/beacon-biosignals/Onda.jl/pull/137. Let me know what you think.

jrevels (approved):

PR #83, opened 2023-04-06
Provide nice error messages when constructing record fields

### LINKED ISSUE #44: Better error messages when passing an object that does not satisfy as schema.
In practice it can be kind of annoying that Legolas doesn't tell you the name of the column that fails to comply with a schema. E.g.

```julia
const Meta = Legolas.@row("ondajobs.meta@1", analysis::String, kind::String, 
                          source::Union{Missing, String}, line::Union{Missing, Int}, 
                          path::String,
                          githash::Union{Missing, String} = OndaJobs.githash())

Meta((;analysis = "test", kind = "a", label = "h", version = v"0.1.0"))
```

Will throw an error like this:

```Julia
ERROR: MethodError: Cannot `convert` an object of type Missing to an object of type String
Closest candidates are:
  convert(::Type{T}, ::Union{InitialValues.SpecificInitialValue{typeof(*)}, InitialValues.SpecificInitialValue{typeof(Base.mul_prod)}}) where T<:Union{AbstractString, Number} at ~/.julia/packages/InitialValues/OWP8V/src/InitialValues.jl:258
  convert(::Type{String}, ::FilePathsBase.AbstractPath) at ~/.julia/packages/FilePathsBase/KUHsP/src/path.jl:117
  convert(::Type{String}, ::String) at ~/.julia/juliaup/julia-1.7.2+0~x64/share/julia/base/essentials.jl:223
  ...
Stacktrace:
 [1] _transform(::Legolas.Schema{Symbol("ondajobs.meta"), 1}; analysis::String, kind::String, source::Missing, line::Missing, path::Missing, githash::Missing, other::Base.Pairs{Symbol, Any, Tuple{Symbol, Symbol}, NamedTuple{(:label, :version), Tuple{String, VersionNumber}}})
   @ OndaJobs ~/Documents/work/beacon/OndaJobs.jl/src/record.jl:157
 [2] transform(schema::Legolas.Schema{Symbol("ondajobs.meta"), 1}; fields::Base.Pairs{Symbol, Any, NTuple{4, Symbol}, NamedTuple{(:analysis, :kind, :label, :version), Tuple{String, String, String, VersionNumber}}})
   @ OndaJobs ~/Documents/work/beacon/OndaJobs.jl/src/record.jl:158
 [3] Legolas.Row(schema::Legolas.Schema{Symbol("ondajobs.meta"), 1}; fields::Base.Pairs{Symbol, Any, NTuple{4, Symbol}, NamedTuple{(:analysis, :kind, :label, :version), Tuple{String, String, String, VersionNumber}}})
   @ Legolas ~/.julia/packages/Legolas/G6pxI/src/rows.jl:201
 [4] Legolas.Row(schema::Legolas.Schema{Symbol("ondajobs.meta"), 1}, fields::NamedTuple{(:analysis, :kind, :label, :version), Tuple{String, String, String, VersionNumber}})
   @ Legolas ~/.julia/packages/Legolas/G6pxI/src/rows.jl:210
 [5] (Legolas.Row{Legolas.Schema{Symbol("ondajobs.meta"), 1}})(args::NamedTuple{(:analysis, :kind, :label, :version), Tuple{String, String, String, VersionNumber}}; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ Legolas ~/.julia/packages/Legolas/G6pxI/src/rows.jl:206
 [6] (Legolas.Row{Legolas.Schema{Symbol("ondajobs.meta"), 1}})(args::NamedTuple{(:analysis, :kind, :label, :version), Tuple{String, String, String, VersionNumber}})
   @ Legolas ~/.julia/packages/Legolas/G6pxI/src/rows.jl:206
 [7] top-level scope
   @ REPL[38]:1
 [8] top-level scope
   @ ~/.julia/packages/Infiltrator/RRI83/src/Infiltrator.jl:567
```

Ideally would throw an error that includes the name of the field that was missed (in this case `path`).

omus:
Fixes #44 but we'll probably want to spawn a new issue for https://github.com/beacon-biosignals/Legolas.jl/issues/44#issuecomment-1292609205. A quick example of this in action:

```julia
julia> using Legolas: Legolas, @schema, @version

julia> @schema "test.field-error" FieldError

julia> function _validate(x)
           x in ("a", "b", "c") || throw(ArgumentError("Must be a, b, or c"))
           return x
       end
_validate (generic function with 1 method)

julia> @version FieldErrorV1 begin
           a::Union{String,Missing} = Legolas.lift(_validate, a)
           b::(<:Union{String,Missing}) = Legolas.lift(_validate, b)
           c::Union{Integer,Missing}
           d::(<:Union{Integer,Missing})
       end

julia> FieldErrorV1(; a="invalid")
ERROR: ArgumentError: Invalid value set for field `a` ("invalid")
Stacktrace:
 [1] FieldErrorV1(; a::String, b::Missing, c::Missing, d::Missing)
   @ Main ~/.julia/dev/Legolas/src/schemas.jl:488
 [2] top-level scope
   @ REPL[5]:1

caused by: ArgumentError: Must be a, b, or c
Stacktrace:
 [1] _validate
   @ ./REPL[3]:2 [inlined]
 [2] lift
   @ ~/.julia/dev/Legolas/src/lift.jl:12 [inlined]
 [3] FieldErrorV1(; a::String, b::Missing, c::Missing, d::Missing)
   @ Main ~/.julia/dev/Legolas/src/schemas.jl:511
 [4] top-level scope
   @ REPL[5]:1
```
I ended up going with this stacked exception approach as it allows for field constructors to easily report issues without having to include the additional context of the name of the field or the value itself.

Update: Modified example to show current exception output

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204322643598304

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -10,14 +10,16 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [compat]
 Arrow = "2"
+Compat = "3.34"
 DataFrames = "1"
 Tables = "1.4"
 julia = "1.6"
 
 [extras]
+Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [targets]
-test = ["Test", "DataFrames", "UUIDs"]
+test = ["Compat", "DataFrames", "Test", "UUIDs"]
--- a/Project.toml
+++ b/Project.toml
@@ -10,7 +10,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [compat]
 Arrow = "2"
-Compat = "3.34"
+Compat = "3.34, 4"
 DataFrames = "1"
 Tables = "1.4"
 julia = "1.6"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Legolas"
 uuid = "741b9549-f6ed-4911-9fbf-4a1c0c97f0cd"
 authors = ["Beacon Biosignals, Inc."]
-version = "0.5.7"
+version = "0.5.8"
 
 [deps]
 Arrow = "69666777-d1a9-59fb-9406-91d4454c9d45"
--- a/src/schemas.jl
+++ b/src/schemas.jl
@@ -483,26 +483,29 @@ function _generate_record_type_definitions(schema_version::SchemaVersion, record
         fdef = :($fname::$T)
         info = get(declared_field_infos, fname, nothing)
         if !isnothing(info)
+            fex = :(throw(ArgumentError("Invalid value set for field $($fsym), expected $($(info.type)), got a value of type $(typeof($fname)) ($(repr($(fname))))")))
             if info.parameterize
                 T = Symbol("_", string(fname, "_T"))
                 push!(type_param_defs, :($T <: $(info.type)))
                 push!(names_of_parameterized_fields, fname)
                 fdef = :($fname::$T)
-                fstmt = :($fname = $(info.statement.args[2]))
                 fstmt = quote
-                    $fname = $(info.statement.args[2])
+                    try
+                        $fname = $(info.statement.args[2])
+                    catch
+                        throw($fex)
+                    end
                     if !($fname isa $(info.type))
                         throw(TypeError($(Base.Meta.quot(record_type_symbol)), "field $($fsym)", $(info.type), $fname))
                     end
-                    # TypeError: in DemoV2, in _a_T, expected _a_T<:String, got Type{SubString{String}}
                 end
             else
                 fstmt = quote
-                    $fname = $(info.statement.args[2])
-                    $fname = try
-                        convert($T, $fname)::$T
-                    catch e
-                        throw(ArgumentError("Invalid value set for field $($fsym): $(repr($(fname)))"))
+                    try
+                        $fname = $(info.statement.args[2])
+                        $fname = convert($T, $fname)::$T
+                    catch
+                        throw($fex)
                     end
                 end
             end
--- a/src/schemas.jl
+++ b/src/schemas.jl
@@ -483,7 +483,13 @@ function _generate_record_type_definitions(schema_version::SchemaVersion, record
         fdef = :($fname::$T)
         info = get(declared_field_infos, fname, nothing)
         if !isnothing(info)
-            fex = :(throw(ArgumentError("Invalid value set for field $($fsym), expected $($(info.type)), got a value of type $(typeof($fname)) ($(repr($(fname))))")))
+            fcatch = quote
+                if $fname isa $(info.type)
+                    throw(ArgumentError("Invalid value set for field $($fsym), custom field assignment failed for: $(repr($(fname)))"))
+                else
+                    throw(ArgumentError("Invalid value set for field $($fsym), expected $($(info.type)), got a value of type $(typeof($fname)) ($(repr($(fname))))"))
+                end
+            end
             if info.parameterize
                 T = gensym(string(fname, "_T"))
                 push!(type_param_defs, :($T <: $(info.type)))
@@ -493,7 +499,7 @@ function _generate_record_type_definitions(schema_version::SchemaVersion, record
                     try
                         $fname = $(info.statement.args[2])
                     catch
-                        throw($fex)
+                        $fcatch
                     end
                     if !($fname isa $(info.type))
                         throw(TypeError($(Base.Meta.quot(record_type_symbol)), "field $($fsym)", $(info.type), $fname))
@@ -505,7 +511,7 @@ function _generate_record_type_definitions(schema_version::SchemaVersion, record
                         $fname = $(info.statement.args[2])
                         $fname = convert($T, $fname)::$T
                     catch
-                        throw($fex)
+                        $fcatch
                     end
                 end
             end
--- a/src/schemas.jl
+++ b/src/schemas.jl
@@ -478,6 +478,7 @@ function _generate_record_type_definitions(schema_version::SchemaVersion, record
     field_definitions = Expr[]
     field_assignments = Expr[]
     for (fname, ftype) in pairs(record_fields)
+        fsym = Base.Meta.quot(fname)
         T = Base.Meta.quot(ftype)
         fdef = :($fname::$T)
         info = get(declared_field_infos, fname, nothing)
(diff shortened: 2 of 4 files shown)


ericphanson:
In the example,
```julia
ERROR: ArgumentError: Invalid value set for field a, expected Union{Missing, String}, got a value of type String ("invalid")
```
feels a little misleading because "expected" seems to imply it got something unexpected, whereas in this case it did get the expected type.

omus:
> In the example,
> 
> ```julia
> ERROR: ArgumentError: Invalid value set for field a, expected Union{Missing, String}, got a value of type String ("invalid")
> ```
> 
> feels a little misleading because "expected" seems to imply it got something unexpected, whereas in this case it did get the expected type.

Reasonable. I can do some revision there.

omus:
I can reproduce the error locally now that I've merged against main

omus:
This PR uncovered that https://github.com/beacon-biosignals/Legolas.jl/pull/82 introduced a very subtle bug where if you are using a parametric schema version and when you attempt to construct a record if the value of a parametric field is not an expected subtype you get an error like this:
```
TypeError: in FieldErrorV1, in 326, expected var\"326\"<:Union{Missing, Integer}, got Type{Float64
```
Instead of the nice ones generated in this PR. I've reverted the change from #82 here as it was just a minor bit of cleanup. I do think I should add a test explicitly for this problem but I'll do that in another PR.

omus:
I looked into the coverage reduction and it appears that the `$fcatch` lines are not being marked as lines which have been tested. As all of the tests I added here are about testing those lines I think we can safely ignore the coverage drop

Project.toml
```diff
@@ -10,14 +10,16 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [compat]
 Arrow = "2"
+Compat = "3.34"
```
kleinschmidt, line 13: can we add a defensive compat 4 compat here too?
```suggestion
Compat = "3.34, 4"
```

src/schemas.jl
```diff
@@ -530,11 +554,13 @@ function _generate_record_type_definitions(schema_version::SchemaVersion, record
                 $(field_assignments...)
                 return new{$(type_param_names...)}($(keys(record_fields)...))
             end
-        end
-        outer_constructor_definitions = quote
             function $R(; $(field_kwargs...))
-                return $R{$((:(typeof($n)) for n in names_of_parameterized_fields)...)}(; $(keys(record_fields)...))
+                $parent_record_application
+                $(field_assignments...)
+                return new{$((:(typeof($n)) for n in names_of_parameterized_fields)...)}($(keys(record_fields)...))
```
kleinschmidt, line 560: why are we switching from kwargs to positional here?

kleinschmidt: oh I see, because we're not falling back to calling `$R` constructor here anymore, makes sense.

test/runtests.jl
```diff
@@ -472,3 +473,95 @@ end
...
+    @testset "field constructor error" begin
+        ex_stack = try
+            FieldErrorV1(; a="invalid")
+        catch
+            current_exceptions()
+        end
+
+        @test length(ex_stack) == 2
+        @test sprint(showerror, ex_stack[1].exception) == "ArgumentError: Must be a, b, or c"
+        @test sprint(showerror, ex_stack[2].exception) == "ArgumentError: Invalid value set for field a (\"invalid\")"
```
kleinschmidt, line 517: this is so nitpicky but can we make the error more clearly set the field name apart?  like `` `a` `` or `:a` or something like that?

src/schemas.jl
```diff
@@ -478,18 +478,42 @@ function _generate_record_type_definitions(schema_version::SchemaVersion, record
...
     field_assignments = Expr[]
     for (fname, ftype) in pairs(record_fields)
+        fsym = Base.Meta.quot(fname)
         T = Base.Meta.quot(ftype)
         fdef = :($fname::$T)
         info = get(declared_field_infos, fname, nothing)
         if !isnothing(info)
+            fcatch = quote
+                if $fname isa $(info.type)
+                    throw(ArgumentError("Invalid value set for field $($fsym) ($(repr($(fname))))"))
```
kleinschmidt, line 488: maybe put the field names in backticks so they stand out?

```suggestion
                    throw(ArgumentError("Invalid value set for field `$($fsym)` ($(repr($(fname))))"))
```

kleinschmidt (approved):
I'm having a biiiiit of trouble following all the macro threads but I think I understand what's going on and it seems fine.  Some extremely nitpicky suggestions on formatting the error messages but otherwise looks good.

src/schemas.jl
omus: There's some real subtle stuff going on here. In #82 I changed this code to no longer be an inner constructor and just make the non-parametric constructor call the parametric constructor. However, with the changes made in this PR it turned out that if a user passed in a value that wasn't a subtype of a specified parametric type we would no longer see the nice error messages and get a very confusing `TypeError` that included the `gensym`d parametric variable which is normally hidden from end users.

I also made this PR https://github.com/beacon-biosignals/Legolas.jl/pull/84 which also makes this change and introduces additional tests specifically to look for this

test/runtests.jl
omus: Ended up chatting about this and we'll go with double quotes. Will call out that the `TypeError` which we also use typically doesn't quote things which is kind of strange:
```julia
julia> Rational{String}
ERROR: TypeError: in Rational, in T, expected T<:Integer, got Type{String}
```

omus: As I was adding the quoting I just noticed Legolas uses backticks for other error messages referring to fields so we'll just stick with that convention

src/schemas.jl
omus: Now that #82 is merged and this PR has merged in those changes these changes no longer exist here :)

omus:
Merging in `main` uncovered and issue with automatic conversion when parametric fields are used. I'll try to fix that

omus:
Fixed the conversion issue but found something else I'll want to address in a follow up:
```julia
julia> using Legolas: Legolas, @schema, @version

julia> @schema "test.demo" Demo

julia> @version DemoV1 begin
           b::(<:Union{Integer,Missing})
       end

julia> DemoV1{Int}(b=1.0)  # Used to fail with the changes in this PR
DemoV1{Int64}: (b = 1,)

julia> DemoV1{Float64}  # Use of gensym makes this ugly
ERROR: TypeError: in DemoV1, in 312, expected var"312"<:Union{Missing, Integer}, got Type{Float64}
Stacktrace:
 [1] top-level scope
   @ REPL[6]:1
```
In #81 when I switched this code to using `gensym` I made this error less readable. Unfortunately as we allow people to write custom code for the field constructors even if we use `_$(field_name_T` as the type parameter it's possible for user code to reassign these parameters.

Personally, I'm inclined to remove the parametric record constructors as I'm not sure when they would ever be used but that is definitely a breaking change we can deal with later.

kleinschmidt:
> Use of gensym makes this ugly

Could you use gensym with the fieldname as a string or something like that to make a slightly more readable generated symbol

kleinschmidt (approved):

omus:
> Could you use gensym with the fieldname as a string or something like that to make a slightly more readable generated symbol

We were using `gensym` and passing in the field name. Something else is going on:
```julia
julia> T = gensym("foo")
Symbol("##foo#312")

julia> ex = quote
           struct Foo{$T <: Integer}
               x::$T
           end
       end
quote
    #= REPL[2]:2 =#
    struct Foo{var"##foo#312" <: Integer}
        #= REPL[2]:3 =#
        x::var"##foo#312"
    end
end

julia> eval(ex)

julia> Foo{String}
ERROR: TypeError: in Foo, in 312, expected var"312"<:Integer, got Type{String}
Stacktrace:
 [1] top-level scope
   @ REPL[4]:1

julia> Meta.dump(ex)
Expr
  head: Symbol block
  args: Array{Any}((2,))
    1: LineNumberNode
      line: Int64 2
      file: Symbol REPL[2]
    2: Expr
      head: Symbol struct
      args: Array{Any}((3,))
        1: Bool false
        2: Expr
          head: Symbol curly
          args: Array{Any}((2,))
            1: Symbol Foo
            2: Expr
              head: Symbol <:
              args: Array{Any}((2,))
                1: Symbol ##foo#312
                2: Symbol Integer
        3: Expr
          head: Symbol block
          args: Array{Any}((2,))
            1: LineNumberNode
              line: Int64 3
              file: Symbol REPL[2]
            2: Expr
              head: Symbol ::
              args: Array{Any}((2,))
                1: Symbol x
                2: Symbol ##foo#312
```

PR #84, opened 2023-04-06
Add parameterized schema version tests

omus:
Adding additional tests around parameterized schema versions as called out in: https://github.com/beacon-biosignals/Legolas.jl/pull/83#issuecomment-1499524213. I also put the tour tests in a separate namespace to ensure that we don't run into schema version collisions.

These tests required that we rollback #82 so that is also included here.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -3,7 +3,10 @@ using Legolas: SchemaVersion, @schema, @version, SchemaVersionDeclarationError, 
 
 @test_throws SchemaVersionDeclarationError("no prior `@schema` declaration found in current module") @version(TestV1, begin x end)
 
-include(joinpath(dirname(@__DIR__), "examples", "tour.jl"))
+# Isolate schema defined in the tour from the rest of the tests
+module Tour
+    include(joinpath(dirname(@__DIR__), "examples", "tour.jl"))
+end
 
 @testset "Legolas.lift" begin
     @test ismissing(Legolas.lift(sin, nothing))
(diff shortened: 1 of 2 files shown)


test/runtests.jl
```diff
@@ -395,6 +404,15 @@ end
...
         @test contains(ds, "Very detailed documentation")
     end
+
+    @testset "parameterized" begin
+        @test typeof(ParamV1(i=1)) === ParamV1{Int}
+        @test typeof(ParamV1{Integer}(i=1)) === ParamV1{Integer}
+        @test typeof(ParamV1{Int}(i=1.0)) === ParamV1{Int}
+        @test_throws TypeError ParamV1{Float64}(i=1)
+        @test_throws TypeError ParamV1(i=1.0)
+        @test_throws InexactError ParamV1{Int}(i=1.1)
```
kleinschmidt, line 414: these all need `;` for the kwarg to be YASG compliant :eye-of-yas:

```suggestion
        @test typeof(ParamV1(; i=1)) === ParamV1{Int}
        @test typeof(ParamV1{Integer}(; i=1)) === ParamV1{Integer}
        @test typeof(ParamV1{Int}(; i=1.0)) === ParamV1{Int}
        @test_throws TypeError ParamV1{Float64}(; i=1)
        @test_throws TypeError ParamV1(; i=1.0)
        @test_throws InexactError ParamV1{Int}(; i=1.1)
```

kleinschmidt (approved):

PR #85, opened 2023-04-08
Print all schema violations in `validate`'s error message

### LINKED ISSUE #56: Print all schema violations in `validate`'s error message
Not just the first one.

xref https://github.com/beacon-biosignals/Legolas.jl/pull/54#discussion_r1005594846

hannahilea:
Close #56. 

```julia
# Setup
using Legolas
using Legolas: SchemaVersion, @schema, @version
using Tables

@schema "foo" Foo
@version FooV1 begin
    a::Vector
    b::AbstractString
    c::Int32
    d::Int32
end
f = FooV1SchemaVersion()
t = Tables.Schema((:a, :b), (Int32, Float64))
Legolas.validate(t, f)
```

`main`:
```julia
julia> Legolas.validate(t, f)
ERROR: ArgumentError: field `a` has unexpected type; expected <:Vector, found Int32
Stacktrace:
 [1] validate(ts::Tables.Schema{(:a, :b), Tuple{Int32, Float64}}, sv::SchemaVersion{:foo, 1})
   @ Legolas ~/Code/Legolas.jl/src/schemas.jl:285
 [2] top-level scope
   @ REPL[8]:1
```

This branch:
```julia
julia> Legolas.validate(t, f)
ERROR: ArgumentError: Tables.Schema violates Legolas schema `foo@1`:
 - Could not find required field: `c`
 - Could not find required field: `d`
 - Incorrect type: `a` expected `<:Vector`, found `Int32`
 - Incorrect type: `b` expected `<:AbstractString`, found `Float64`
Provided Tables.Schema:
 :a  Int32
 :b  Float64
```
(Note: was updated following review feedback below)

---------

Punch list: 
- [x] capture for multiple violations
- [x] show multiple violations in validation script
  - [x] example old version
  - [x] example new version
- [x] add/update tests
- [x] bump version


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Legolas"
 uuid = "741b9549-f6ed-4911-9fbf-4a1c0c97f0cd"
 authors = ["Beacon Biosignals, Inc."]
-version = "0.5.8"
+version = "0.5.9"
```
hannahilea, line 4: non-breaking: existing functions weren't removed; user-facing change is body of error messages, although they are still the same type of error (`ArgumentError`).

src/schemas.jl
```diff
@@ -254,35 +254,79 @@ accepted_field_type(::SchemaVersion, ::Type{Symbol}) = Union{Symbol,String}
...
+- If `f::T` is present in `ts`, ensure that `T <: A` or else append `f::Symbol => T::DataType` to output vector.
+- If `f` isn't present in `ts`, ensure that `Missing <: A` or else append `f::Symbol => missing::Missing` to output vector.
+
+Otherwise, return `Any[]`.
+
+See also: [`Legolas.find_violation`](@ref), [`Legolas.validate`](@ref), [`Legolas.complies_with`](@ref)
+"""
+find_violations(::Tables.Schema, sv::SchemaVersion) = throw(UnknownSchemaVersionError(sv))
+
+function _find_violations end
```
hannahilea, line 286: tbh I'm not sure what the role of this is---I added it because I saw there was a `function _find_violation end` above. Thoughts on why the above one exists (and therefore whether this one should)?

test/runtests.jl
```diff
@@ -316,16 +316,44 @@ end
...
         t = Tables.Schema((:a, :y, :z), (Int32, String, Any))
         for s in (GrandchildV1SchemaVersion(), ChildV1SchemaVersion(), ParentV1SchemaVersion())
-            @test_throws ArgumentError("could not find expected field `x` in $t") Legolas.validate(t, s)
+            @test_throws ArgumentError("Could not find expected field `x` in $t") Legolas.validate(t, s)
             @test !Legolas.complies_with(t, s)
             @test isequal(Legolas.find_violation(t, s), :x => missing)
+            @test isequal(Legolas.find_violations(t, s), [:x => missing])
+        end
+
+        t = Tables.Schema((:a, :z), (Int32, Any))
```
hannahilea, line 325:
```suggestion
        # Multiple expected field violations
        t = Tables.Schema((:a, :z), (Int32, Any))
```

src/schemas.jl
```diff
@@ -254,35 +254,79 @@ accepted_field_type(::SchemaVersion, ::Type{Symbol}) = Union{Symbol,String}
...
 function _find_violation end
 
+"""
+    Legolas.find_violations(ts::Tables.Schema, sv::Legolas.SchemaVersion)
+
+Return vector of all schema violations. For required fields `f::F` of `sv`:
+
+- Define `A = Legolas.accepted_field_type(sv, F)`
+- If `f::T` is present in `ts`, ensure that `T <: A` or else append `f::Symbol => T::DataType` to output vector.
+- If `f` isn't present in `ts`, ensure that `Missing <: A` or else append `f::Symbol => missing::Missing` to output vector.
```
glennmoy, line 278: I'm a little confused by the wording of this...is it being imperative and asking the user to define this function or describing what the function is doing?

src/schemas.jl
```diff
@@ -449,11 +493,31 @@ function _generate_validation_definitions(schema_version::SchemaVersion)
             isnothing(result) || return $fname => result
         end)
     end
+
+    multifield_violation_check = Expr[] # the function body
+    myvector = gensym()
```
glennmoy, line 498: I guess this is to avoid collision with pre-defined variables?

src/schemas.jl
```diff
@@ -449,11 +493,31 @@ function _generate_validation_definitions(schema_version::SchemaVersion)
             isnothing(result) || return $fname => result
```
glennmoy, line 493: this seems to be looping over all violations so I guess the issue here is that when the  `return` is evaluated it exits the loop early and so doesn't show all results?

glennmoy: played around a MWE locally and this seems to be the case

src/schemas.jl
```diff
@@ -254,35 +254,79 @@ accepted_field_type(::SchemaVersion, ::Type{Symbol}) = Union{Symbol,String}
...
+            push!(field_err, field)
+        else
+            expected = getfield(required_fields(sv), field)
+            push!(type_err, (field, expected, violation))
+        end
+    end
+    err = ""
+    if !isempty(field_err)
+        fields = "`" * join(field_err, "`, `") * "`"
+        err *= string("Could not find expected field", length(field_err) > 1 ? "s " : " ", fields, " in $ts")
```
glennmoy, line 314:
```suggestion
        err *= string("Could not find expected field(s)",  fields, " in $ts")
```

FWIW I'd be happy enough with this and would find it easier to read as well

src/schemas.jl
```diff
@@ -254,35 +254,79 @@ accepted_field_type(::SchemaVersion, ::Type{Symbol}) = Union{Symbol,String}
...
+    if length(type_err) == 1
+        (field, expected, violation) = only(type_err)
+        err *= "Field `$field` has unexpected type; expected <:$expected, found $violation"
+    elseif length(type_err) > 1
+        err *= "Fields have unexpected type:\n"
+        for (field, expected, violation) in type_err
+            err *= " `$field`: expected <:$expected, found $violation\n"
+        end
+    end
+    throw(ArgumentError(err))
```
glennmoy, line 329: I'm finding it a bit difficult to follow the error message construction with all of the nested if statements and conditional concatenations...

I wonder if this could be made simpler to read and reason about by being simply additive with each violation found?, e.g.
```julia
err_msg = "Schema violations found:\n"
for err in field_err
    err_msg *= "Missing field: $err\n"
end
for (field, expected, violation) in type_err
    err_msg *= "Incorrect type: `$field` expected `<:$expected`, found $violation \n"
end
throw(ArgumentError(err_msg))
```

(we could even roll this into the loop over `results` above - and remove the needing for constructing the `type_err` or `field_err` vectors - but then we might get a mix of error messages which wouldn't look as neat...)

glennmoy (commented):
I wonder if we could now delete the code for `find_violation` and simply overload it with `find_violation(x...) = first(find_violations(x...))` ? 

Or even better to `@deprecate` it  with the view to removing it altogether in the next breaking release!

Having two functions do basically the same thing but only differ semantically has potential for headaches down the line...in particular if they started to diverge functionally or one introduced a bug, etc.

src/schemas.jl
glennmoy: seems to be describing what is being done behind the scenes when constructing a schema

hannahilea: Yeah---so I adapted this from the description in the singular docstring, even though I think both of them could/should be shorter. I'll update.

src/schemas.jl
hannahilea: yeah, exactly---which I suspect could make for significantly faster validation for cases where complete, user-readable errors are less important (if the various schemas are huge and/or complex)? I think that's perhaps a reason not do do:

> I wonder if we could now delete the code for find_violation and simply overload it with find_violation(x...) = first(find_violations(x...)) ?
> And maybe even @deprecate it with the view to removing it altogether in the next breaking release!

For example, `complies_with` uses this `find_violation` (singular!) approach. I _would_ like to rename this "`find_first_violation`" or something, but am not sure it's worth the breaking change? Or maybe it's worth following your proposed deprecation path but for the rename...

> Having two functions do basically the same thing but only differ semantically has potential for headaches down the line...in particular if they started to diverge functionally or one introduced a bug, etc

Yeah, this is a good point. It would be nicer if the macros that generate the two shared code---and tbh, the reason that they don't is that I'm so new to this macro-on-macro style that it was tricky enough for me to figure out how do do them separately! I'm not _too_ worried that they diverge yet, given that we have tests for both of them, and they're close enough in code location that it seems likely that someone adjusting one would see the other as well, but this is a great point. Let me do a quick timebox to attempt sharing code, now that I understand this code better---if I can't figure something out quickly, I think we could leave it for the future (or someone else could propose a fix here! :) ).

src/schemas.jl
hannahilea: I like this! Thanks, updating

hannahilea: done! (slightly modified, see updated example at top of PR :) )

src/schemas.jl
hannahilea: Right okay, refactored! Not sure it's easier to read, but at least they share a single codepath now...?

Does anyone have strong opinions re: the rename/deprecation?

src/schemas.jl
```diff
@@ -440,18 +462,38 @@ function _generate_schema_version_definitions(schema_version::SchemaVersion, par
...
+                if !isnothing(result) && $(fail_fast)
+                    return $fname => result
+                elseif !isnothing(result)
+                    push!($myvector, ($fname => result))
+                end
+            end)
+        end
+        push!(violations, quote
+            return $(fail_fast) ? nothing : $myvector
         end)
```
glennmoy, line 484: not sure I get this part...am I correct to say is this supposed to be the final return value from `find_violation(s)`? 

Would it be better to express the return value inside those functions explicitly?  
As it stands they both seem to `return nothing` which is misleading...

src/schemas.jl
glennmoy: Thanks for refactoring - I think sharing code is definitely a positive step!

> I suspect could make for significantly faster validation for cases where complete, user-readable errors are less important (if the various schemas are huge and/or complex)?

Hmm...how large/complex do these schemas get? And would the cost of exhaustively running `_check_for_expected_field` really be significant?

I won't let this bike-shedding block the PR but I would be surprised to learn if it really was a computational bottleneck! 
If not, it would be trivial to switch `complies_with` to check `isempty(find_violations(...))`

src/schemas.jl
```diff
@@ -440,18 +462,38 @@ function _generate_schema_version_definitions(schema_version::SchemaVersion, par
...
-            $(field_violation_check_statements...)
+            $(_violation_check(; fail_fast=true)...)
+            return nothing
+        end
+
+        # Multiple violation reporting
+        function $(Legolas).find_violations(ts::$(Tables).Schema, sv::$(Base.Meta.quot(typeof(schema_version))))
+            $(_violation_check(; fail_fast=false)...)
             return nothing
         end
```
omus, line 498: IMO we should use the code from `_violation_check` and generate a single function `_find_violations` which is then called by:
```julia
find_violation(ts::Tables.Schema, sv::SchemaVersion) = _find_violations(ts, vs; fail_fast=true)
find_violations(ts::Tables.Schema, sv::SchemaVersion) = _find_violations(ts, vs; fail_fast=false)
```

src/schemas.jl
omus: I totally agree

src/schemas.jl
```diff
@@ -254,35 +254,57 @@ accepted_field_type(::SchemaVersion, ::Type{Symbol}) = Union{Symbol,String}
...
+        if ismissing(violation)
+            push!(field_err, field)
+        else
+            expected = getfield(required_fields(sv), field)
+            push!(type_err, (field, expected, violation))
+        end
+    end
+    err_msg = "Schema violation(s) found for $ts\n"
+    for err in field_err
+        err_msg *= " - Could not find expected field: `$err`\n"
```
omus, line 302:
```suggestion
        err_msg *= " - Could not find required field: `$err`\n"
```

src/schemas.jl
```diff
@@ -254,35 +254,57 @@ accepted_field_type(::SchemaVersion, ::Type{Symbol}) = Union{Symbol,String}
...
-    isnothing(result) && return nothing
-    field, violation = result
-    ismissing(violation) && throw(ArgumentError("could not find expected field `$field` in $ts"))
-    expected = getfield(required_fields(sv), field)
-    throw(ArgumentError("field `$field` has unexpected type; expected <:$expected, found $violation"))
+    results = find_violations(ts, sv)
+    isempty(results) && return nothing
+
+    field_err = Any[]
+    type_err = Any[]
```
omus, line 290:
```suggestion
    field_err = Symbol[]
    type_err = Tuple{Symbol,Type,Type}[]
```

src/schemas.jl
omus: Here's a first pass at the `_violation_check` function:
```julia
SV = Base.Meta.quot(typeof(schema_version))
quote
    function _violation_check(ts::$(Tables).Schema, sv::$SV; fail_fast::Bool)
        violations = Pair{Symbol,Union{Type,Missing}}[]
        for (fname, ftype) in pairs(required_fields(sv))
            S = Legolas.accepted_field_type(sv, ftype)
            result = Legolas._check_for_expected_field(ts, fname, S)
            if !isnothing(result)
                push!(violations, fname => result)
                fail_fast && break
            end
        end
        return violations
    end
end
```

src/schemas.jl
```diff
@@ -254,35 +254,57 @@ accepted_field_type(::SchemaVersion, ::Type{Symbol}) = Union{Symbol,String}
...
-- Define `A = Legolas.accepted_field_type(sv, F)`
-- If `f::T` is present in `ts`, ensure that `T <: A` or else immediately return `f::Symbol => T::DataType`.
-- If `f` isn't present in `ts`, ensure that `Missing <: A` or else immediately return `f::Symbol => missing::Missing`.
-
-Otherwise, return `nothing`.
-
-See also: [`Legolas.validate`](@ref), [`Legolas.complies_with`](@ref)
+See also: [`Legolas.find_violations`](@ref), [`Legolas.validate`](@ref), [`Legolas.complies_with`](@ref)
 """
 find_violation(::Tables.Schema, sv::SchemaVersion) = throw(UnknownSchemaVersionError(sv))
```
omus, line 262: I think we should deprecate this function in favor of `find_violations`

test/runtests.jl
```diff
@@ -316,16 +316,64 @@ end
...
         for s in (GrandchildV1SchemaVersion(), ChildV1SchemaVersion(), ParentV1SchemaVersion())
-            @test_throws ArgumentError("field `x` has unexpected type; expected <:$(Vector), found $(Complex{Float64})") Legolas.validate(t, s)
+            msg = """
+                  Schema violation(s) found for Tables.Schema:
+                   :a  Int32
+                   :z  Any
+                   - Could not find expected field: `x`
+                   - Could not find expected field: `y`
+                  """
```
omus, line 341: Looking at this error message in isolation I think we could provide more context yet:
```suggestion
            msg = """
                  Tables.Schema violates Legolas schema $s:
                    - Could not find expected field: `x`
                    - Could not find expected field: `y`
                  Provided Tables.Schema:
                   :a  Int32
                   :z  Any
                  """
```

hannahilea: updated, and updated top level PR comment to reflect new format.

src/schemas.jl
hannahilea: so there actually is a difference in what gets returned by these two functions---find_violation returns `nothing` or the first violation, where `find_violations` returns the whole (potentially empty) list of violations. Since we're planning to deprecate `find_violation` anyway, I'd like to just use what's here already and then remove the extraneous lines when when we remove find_violations outright. (I attempted to :just: pull this chunk in, but it didn't work out of the box, and I don't think the priority is high enough to spend more time fighting it.)

src/schemas.jl
```diff
@@ -254,35 +254,58 @@ accepted_field_type(::SchemaVersion, ::Type{Symbol}) = Union{Symbol,String}
...
+            push!(type_err, (field, expected, violation))
+        end
+    end
+    err_msg = "Tables.Schema violates Legolas schema `$(string(name(sv), "@", version(sv)))`:\n"
+    for err in field_err
+        err_msg *= "  - Could not find required field: `$err`\n"
+    end
+    for (field, expected, violation) in type_err
+        err_msg *= "  - Incorrect type: `$field` expected `<:$expected`, found `$violation`\n"
+    end
```
omus, line 306: We should probably mirror the indentation level used by `Table.Schema`
```suggestion
    for err in field_err
        err_msg *= " - Could not find required field: `$err`\n"
    end
    for (field, expected, violation) in type_err
        err_msg *= " - Incorrect type: `$field` expected `<:$expected`, found `$violation`\n"
    end
```

src/schemas.jl
omus: @hannahilea I was intending for you to the above in conjunction with these:
```julia
function Legolas.find_violation(ts::Tables.Schema, sv::SchemaVersion)
    violations = _violation_check(sv; fail_fast=true)
    return isempty(violations) ? nothing : first(violations)
end

function Legolas.find_violations(ts::Tables.Schema, sv::SchemaVersion)
    return _violation_check(sv; fail_fast=false)
end
```
I do think this is important to make this change to help transition this package into a more maintainable state by others.

src/schemas.jl
hannahilea: done!

src/schemas.jl
hannahilea: I agree that this would be nice, but :just: copying in the changes here did not successfully run for me; I think there's probably something basic that I'm neglecting to also add or something. A little bit of playing around didn't get me there, and I don't have any more time to allocate to this side-quest 😅 .  Since 

> to help transition this package into a more maintainable state by others

is not the current goal, and I don't think the changes are _less_ readable than before, I'd like to merge it as is and do any readability changes in future PRs.

hannahilea: Unrelated note for future readers: my above statement that 
> so there actually is a difference in what gets returned by these two functions---find_violation returns nothing or the first violation, where find_violations returns the whole (potentially empty) list of violations

is now incorrect---I updated it so that `find_violations` _also_ returns `nothing` when no violations are found, rather than an empty list. That's more in line with our other validation-type functions, and will make deprecation fixing easier for anyone who depends on this.

examples/tour.jl
```diff
@@ -71,26 +71,26 @@ for s in [Tables.Schema((:a, :b, :c, :d), (Real, String, Any, AbstractVector)),
...
+    # if `validate` finds any violations, it throws an error indicating the violations;
     # returns `nothing` otherwise
     @test validate(s, FooV1SchemaVersion()) isa Nothing
 
-    # if `find_violation` finds a violation, it returns a tuple indicating the relevant
-    # field and its violation; returns `nothing` otherwise
-    @test isnothing(find_violation(s, FooV1SchemaVersion()))
+    # if `find_violations` finds any violations, it returns a vector of tuples indicating the relevant
+    # fields and their violations; returns `nothing` otherwise
+    @test isnothing(find_violations(s, FooV1SchemaVersion()))
```
omus, line 80: I think the new `find_violations` function should always return a vector:
```suggestion
    # if `find_violations` finds any violations, it returns a vector of tuples indicating the relevant
    # fields and their violations
    @test isempty(find_violations(s, FooV1SchemaVersion()))
```
I made this change in https://github.com/beacon-biosignals/Legolas.jl/pull/87 along with another change if you want to pull it in

src/schemas.jl
hannahilea: merged in @omus's suggestions via #87

src/schemas.jl
```diff
@@ -440,19 +474,18 @@ function _generate_schema_version_definitions(schema_version::SchemaVersion, par
...
-        push!(field_violation_check_statements, quote
-            S = $Legolas.accepted_field_type(sv, $ftype)
-            result = $Legolas._check_for_expected_field(ts, $fname, S)
-            isnothing(result) || return $fname => result
-        end)
-    end
+    SV = Base.Meta.quot(typeof(schema_version))
     return quote
-        function $(Legolas).find_violation(ts::$(Tables).Schema, sv::$(Base.Meta.quot(typeof(schema_version))))
-            $(field_violation_check_statements...)
```
jrevels, line 454: i'd actually keep `field_violation_check_statements` and the original inlined generated definition for `find_violation`, which is nicely non-allocating / optimizable

then, to approach defining `find_violations`, i'd construct something like `field_violation_push_statements` (or w/e, probably a better name exists lol) in the same loop that constructs `field_violation_check_statements`; very DRYable since the only difference is `push!(violations, x)` instead of `return x` (where `x` is `$fname => result`) 

then generate the `find_violations` definition similar to the singular version

src/schemas.jl
```diff
@@ -251,48 +251,82 @@ otherwise constitutes type piracy and should be avoided.
...
-For required field `f::F` of `sv`:
+Return first schema violation; otherwise, return `nothing`. For description of
+schema violations, see [`Legolas.find_violations`](@ref).
 
-- Define `A = Legolas.accepted_field_type(sv, F)`
-- If `f::T` is present in `ts`, ensure that `T <: A` or else immediately return `f::Symbol => T::DataType`.
-- If `f` isn't present in `ts`, ensure that `Missing <: A` or else immediately return `f::Symbol => missing::Missing`.
+See also: [`Legolas.find_violations`](@ref), [`Legolas.validate`](@ref), [`Legolas.complies_with`](@ref)
+"""
+function find_violation(ts::Tables.Schema, sv::SchemaVersion)
```
jrevels, line 264: same as https://github.com/beacon-biosignals/Legolas.jl/pull/85/files#r1169390493

but i'd not deprecate this and instead keep the original more optimal short-circuiting implementation

(this method is useful in scenarios where you have small record batches being communicated and you want the common "this batch is valid" case at the communication boundary to be fast)

src/schemas.jl
```diff
@@ -251,48 +251,82 @@ otherwise constitutes type piracy and should be avoided.
...
     Legolas.find_violation(ts::Tables.Schema, sv::Legolas.SchemaVersion)
 
-For required field `f::F` of `sv`:
+Return first schema violation; otherwise, return `nothing`. For description of
+schema violations, see [`Legolas.find_violations`](@ref).
 
-- Define `A = Legolas.accepted_field_type(sv, F)`
-- If `f::T` is present in `ts`, ensure that `T <: A` or else immediately return `f::Symbol => T::DataType`.
-- If `f` isn't present in `ts`, ensure that `Missing <: A` or else immediately return `f::Symbol => missing::Missing`.
+See also: [`Legolas.find_violations`](@ref), [`Legolas.validate`](@ref), [`Legolas.complies_with`](@ref)
```
jrevels, line 262: this docstring specification of what constitutes a "violation" should be kept as-is - it's an important part of how Legolas operates

it's also IMO a bit easier to parse and more precise than the more prose-y version added in `find_violations` (and makes the connection to `accepted_field_type` clear/obvious)

src/schemas.jl
hannahilea: 🙃 https://github.com/beacon-biosignals/Legolas.jl/pull/85#discussion_r1166000126 & #87 

I will revert these!

src/schemas.jl
hannahilea: done

src/schemas.jl
hannahilea: done!

src/schemas.jl
hannahilea: Done

glennmoy (approved):

src/schemas.jl
glennmoy: >this method is useful in scenarios where you have small record batches being communicated and you want the common "this batch is valid" case at the communication boundary to be fast

@jrevels can you elaborate a bit further on this? I'm curious to know what use case this is

src/schemas.jl
```diff
@@ -440,19 +473,37 @@ function _generate_schema_version_definitions(schema_version::SchemaVersion, par
...
+                if !isnothing(result) && $(fail_fast)
+                    return $fname => result
+                elseif !isnothing(result)
+                    push!($myvector, ($fname => result))
+                end
+            end)
+        end
+        push!(violations, quote
+            return $(fail_fast) ? nothing : $myvector
         end)
```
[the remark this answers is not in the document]
jrevels: will commit @omus 's suggestion (which i like) + my own lil stylistic follow-up

src/schemas.jl
```diff
@@ -440,19 +473,43 @@ function _generate_schema_version_definitions(schema_version::SchemaVersion, par
...
+        push!(statements, if fail_fast
+            :(return nothing)
+        else
+            :(return $violations)
         end)
+        return statements
+    end
+        return violations
     end
+
```
jrevels, line 504:
```suggestion
    # When `fail_fast == true`, return first violation found rather than all violations
    _violation_check = (; fail_fast::Bool) -> begin
        statements = Expr[]
        violations = gensym()
        fail_fast || push!(statements, :($violations = Pair{Symbol,Union{Type,Missing}}[]))
        for (fname, ftype) in pairs(required_fields(schema_version))
            fname = Base.Meta.quot(fname)
            found = :($fname => result)
            handle_found = fail_fast ? :(return $found) : :(push!($violations, $found))
            push!(statements, quote
                S = $Legolas.accepted_field_type(sv, $ftype)
                result = $Legolas._check_for_expected_field(ts, $fname, S)
                isnothing(result) || $handle_found
            end)
        end
        push!(statements, :(return $(fail_fast ? :nothing : violations)))
        return statements
    end
```

src/schemas.jl
```diff
@@ -262,35 +262,68 @@ For required field `f::F` of `sv`:
...
+Return vector of all violations, such that for required field `f::F` of `sv`:
+
+- Define `A = Legolas.accepted_field_type(sv, F)`
+- If `f::T` is present in `ts`, ensure that `T <: A` or else append `f::Symbol => T::DataType`.
+- If `f` isn't present in `ts`, ensure that `Missing <: A` or else append `f::Symbol => missing::Missing`.
+
+Otherwise, return `Pair{Symbol,Union{Type,Missing}}[]`.
+
+To immediately return the first violation found, use [`Legolas.find_violation`](@ref).
```
jrevels, line 282:
```suggestion
Return a `Vector{Pair{Symbol,Union{Type,Missing}}}` of all of `ts`'s violations with respect to `sv`.

This function's notion of "violation" is defined by [`Legolas.find_violation`](@ref), which immediately returns the first violation found; prefer to use that function instead of `find_violations` in situations where you only need to detect *any* violation instead of *all* violations. 
```

jrevels (approved):
LGTM! made a few commits to get across finish line
 
 can we add the new `Legolas.find_violations` function to the API documentation at `docs/src/index.md` before merging? but feel free to merge after that + CI passes!

PR #86, opened 2023-04-17
Add CODEOWNERS

omus:
Adding @jrevels as the code owner for this repo.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+*   @jrevels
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -0,0 +1 @@
+*   @jrevels
```
jrevels, line 1:
```suggestion
*   @jrevels @omus 
```

all for one and one for all

PR #87, opened 2023-04-17
Use `find_violations` in `find_violation`

omus:
Updates #85 to generate only `find_violations` methods. Additionally, `find_violations` also returns an empty array when no violations around found, the deprecation now provides guidance on how to update code, and I dropped `fast_fail` as this code isn't performance critical and if it is we may want to return a generator instead.

### DIFF
--- a/examples/tour.jl
+++ b/examples/tour.jl
@@ -76,8 +76,8 @@ for s in [Tables.Schema((:a, :b, :c, :d), (Real, String, Any, AbstractVector)), 
     @test validate(s, FooV1SchemaVersion()) isa Nothing
 
     # if `find_violations` finds any violations, it returns a vector of tuples indicating the relevant
-    # fields and their violations; returns `nothing` otherwise
-    @test isnothing(find_violations(s, FooV1SchemaVersion()))
+    # fields and their violations
+    @test isempty(find_violations(s, FooV1SchemaVersion()))
 end
 
 # ...while the below `Tables.Schema`s do not:
--- a/src/schemas.jl
+++ b/src/schemas.jl
@@ -263,8 +263,8 @@ See also: [`Legolas.find_violations`](@ref), [`Legolas.validate`](@ref), [`Legol
 """
 function find_violation(ts::Tables.Schema, sv::SchemaVersion)
     Base.depwarn(
-        "The `find_violation` function will be deprecated in a future release. " *
-        "Please use `find_violations(args...)` instead of `find_violation(args...)`.",
+        "`find_violation(ts, sv)` is deprecated, use: `let v = find_violations(ts, sv); " *
+        "isempty(v) ? nothing : v; end` instead.",
         :find_violation,
     )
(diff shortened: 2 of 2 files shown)


src/schemas.jl
```diff
@@ -464,41 +476,19 @@ end
...
+            violations = Pair{Symbol,Union{Type,Missing}}[]
+            for (fname, ftype) in pairs(Legolas.required_fields(sv))
+                S = Legolas.accepted_field_type(sv, ftype)
+                result = Legolas._check_for_expected_field(ts, fname, S)
+                if !isnothing(result)
+                    push!(violations, fname => result)
+                    fail_fast && break
+                end
+            end
+            return violations
```
hannahilea, line 491:
```suggestion
            return isempty(violations) ? nothing : violations
```

examples/tour.jl
```diff
@@ -76,8 +76,8 @@ for s in [Tables.Schema((:a, :b, :c, :d), (Real, String, Any, AbstractVector)),
     @test validate(s, FooV1SchemaVersion()) isa Nothing
 
     # if `find_violations` finds any violations, it returns a vector of tuples indicating the relevant
-    # fields and their violations; returns `nothing` otherwise
-    @test isnothing(find_violations(s, FooV1SchemaVersion()))
+    # fields and their violations
+    @test isempty(find_violations(s, FooV1SchemaVersion()))
```
hannahilea, line 80: https://github.com/beacon-biosignals/Legolas.jl/pull/85#discussion_r1169236443

these should still be `nothing`

omus: > these should still be `nothing`

Why though? `find_violations` is a new function added in #85 so we can have it use an empty array instead of `nothing` which seems better. As long as we don't modify `find_violation`'s behavior nothing is broken

omus: > > these should still be `nothing`
> 
> Why though? `find_violations` is a new function added in #85 so we can have it return an empty array instead of `nothing` which seems better. As long as we don't modify `find_violation`'s behavior nothing is broken

src/schemas.jl
```diff
@@ -259,21 +261,30 @@ schema violations, see [`Legolas.find_violations`](@ref).
 
 See also: [`Legolas.find_violations`](@ref), [`Legolas.validate`](@ref), [`Legolas.complies_with`](@ref)
 """
-find_violation(::Tables.Schema, sv::SchemaVersion) = throw(UnknownSchemaVersionError(sv))
+function find_violation(ts::Tables.Schema, sv::SchemaVersion)
+    Base.depwarn(
+        "`find_violation(ts, sv)` is deprecated, use: `let v = find_violations(ts, sv); " *
+        "isempty(v) ? nothing : v; end` instead.",
+        :find_violation,
+    )
```
omus, line 269: Could use `Base.@deprecate` here but the expression it prints ends up expanding the ternary expression which makes this more verbose than I intended

hannahilea (approved):

PR #88, opened 2023-06-06
Add `workflow_dispatch` trigger to CI

hannahilea:

PR #89, opened 2023-06-06
Add `accepted_field_type` for abstract vectors

hannahilea:
ref https://github.com/apache/arrow-julia/issues/452

Add `accepted_field_type`s as proposed by @ericphanson in above issue. No additional tests added BUT `main` is currently failing when run on latest (v2.6) Arrow.jl (e.g. in #88), and no longer fails with this fix.

### DIFF
--- a/src/schemas.jl
+++ b/src/schemas.jl
@@ -242,6 +242,9 @@ Legolas itself defines the following default overloads:
     accepted_field_type(::SchemaVersion, T::Type) = T
     accepted_field_type(::SchemaVersion, ::Type{UUID}) = Union{UUID,UInt128}
     accepted_field_type(::SchemaVersion, ::Type{Symbol}) = Union{Symbol,String}
+    accepted_field_type(::SchemaVersion, ::Type{Vector{String}}) = AbstractVector{<:AbstractString}
+    accepted_field_type(::SchemaVersion, ::Type{Vector{T}}) where T = AbstractVector{<:T}
+    accepted_field_type(::SchemaVersion, ::Type{Vector}) = AbstractVector
 
 Outside of these default overloads, this function should only be overloaded against specific
 `SchemaVersion`s that are authored within the same module as the overload definition; to do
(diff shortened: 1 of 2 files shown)


ericphanson:
Maybe we should also file an issue to add more tests for this as a followup, to cover the changed lines and ensure we don't regress

src/schemas.jl
```diff
@@ -250,6 +253,9 @@ otherwise constitutes type piracy and should be avoided.
 @inline accepted_field_type(::SchemaVersion, T::Type) = T
 accepted_field_type(::SchemaVersion, ::Type{UUID}) = Union{UUID,UInt128}
 accepted_field_type(::SchemaVersion, ::Type{Symbol}) = Union{Symbol,String}
+accepted_field_type(::SchemaVersion, ::Type{Vector{String}}) = AbstractVector{<:AbstractString}
```
jrevels, line 256: if this were true we'd want to define:

```
accepted_field_type(::SchemaVersion, ::Type{String}) = AbstractString
```

and e.g. make sure the `Symbol` definition above maps to `AbstractString` not `String`, etc.

src/schemas.jl
```diff
@@ -250,6 +253,9 @@ otherwise constitutes type piracy and should be avoided.
...
+accepted_field_type(::SchemaVersion, ::Type{Vector{T}}) where T = AbstractVector{<:T}
```
jrevels, line 257:
```suggestion
accepted_field_type(sv::SchemaVersion, ::Type{<:Vector{T}}) where T = AbstractVector{<:(accepted_field_type(sv, T))}
```

?

i think this should also obviate the need for the other `accepted_field_type` definitions for `Vector` here

hannahilea: When `accepted_field_type(::SchemaVersion, ::Type{Vector}) = AbstractVector` is removed, we still hit error in [this test](https://github.com/beacon-biosignals/Legolas.jl/blob/321b19c88c5c7e07c616278f2aaf876495533667/test/runtests.jl#L468):
```julia
┌ Warning: The `Tables.Schema` of the `Arrow.Table` read via `Legolas.read(io_or_path)` does not appear to
│ comply with the `Legolas.SchemaVersion` indicated by the table's metadata (`SchemaVersion("test.parent", 1)`). Try invoking
│ `Legolas.read(io_or_path; validate=false)` to inspect the table.
└ @ Legolas ~/Code/Legolas.jl/src/tables.jl:171
miscellaneous Legolas/src/tables.jl tests: Error During Test at /Users/hrobertson/Code/Legolas.jl/test/runtests.jl:468
  Test threw exception
  Expression: t == [NamedTuple(ParentV1(r)) for r = Tables.rows(Legolas.read(path))]
  ArgumentError: Tables.Schema violates Legolas schema `test.parent@1`:
   - Incorrect type: `x` expected `<:Vector`, found `SubArray{Int64, 1, Arrow.Primitive{Int64, Vector{Int64}}, Tuple{UnitRange{Int64}}, true}`
  Provided Tables.Schema:
   :x  SubArray{Int64, 1, Arrow.Primitive{Int64, Vector{Int64}}, Tuple{UnitRange{Int64}}, true}
   :y  String
  Stacktrace:
   [1] validate(ts::Tables.Schema{(:x, :y), Tuple{SubArray{Int64, 1, Arrow.Primitive{Int64, Vector{Int64}}, Tuple{UnitRange{Int64}}, true}, String}}, sv::SchemaVersion{Symbol("test.parent"), 1})
```

jrevels: yeah, have to keep that one actually, my bad

jrevels (approved):

Issue #63, opened 2022-11-03
add union-splitting fallback for `accepted_field_type`


jrevels:
```jl
julia> Legolas.@schema "example" Example

julia> Legolas.@version ExampleV1 begin id::Union{UUID,Missing} end

julia> Legolas.accepted_field_type(ExampleV1SchemaVersion(), UUID)
Union{UUID, UInt128}

julia> Legolas.accepted_field_type(ExampleV1SchemaVersion(), Union{UUID,Missing})
Union{Missing, UUID}
```

I have to explicitly overload `accepted_field_type` for `Union{UUID,Missing}`. 

Though it's not incorrect at all, needing to do this is annoying and unintuitive

It'd be nice to have an `accepted_field_type` fallback that automatically splits union types, applies the `accepted_field_type` transform to each constituent type, and then re-unions the constituents

jrevels:
I wonder whether it's also worth implementing a similar fallback for eltypes of Arrow.jl-supported container types (`Vector`, `Tuple`, etc.) which basically exhibit this same issue

jrevels:
@omus called out (or, i inferred some code he wrote 😁 ) that `invoke` may also be useful here in some cases, e.g.:

```
function accepted_field_type(sv::Legolas.SchemaVersion, ::Type{T}) where T >: Missing
     A = accepted_field_type(sv, nonmissingtype(T))
     return Union{A,Missing}
end
```

the fully generic one might require general union-splitting though

iamed2:
Julia 1.9+ now has `typesplit` which could be used for generic union-splitting (not exported unfortunately, but better than `Core.Compiler.typesubtract`)

PR #96, opened 2023-06-13
Add `accepted_field_type` for `Union`s with `Missing`

ararslan:
PR #89 fixed compatibility with Arrow v2.6 except when the schema field's declared type is e.g. `Union{Vector{T},Missing}` and what's passed for validation is some `AbstractVector`. This adds a method to `accepted_field_type` that recurses on the non-`Missing` type in a `Union` with `Missing`. Note that this also necessitated adding an explicit `accepted_field_type` method for `Any` to avoid infinite recursion with the fallback definition, since `(Union{T,Missing} where {T}) == Any`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Legolas"
 uuid = "741b9549-f6ed-4911-9fbf-4a1c0c97f0cd"
 authors = ["Beacon Biosignals, Inc."]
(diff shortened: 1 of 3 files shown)


jrevels (approved):
xref https://github.com/beacon-biosignals/Legolas.jl/issues/63

PR #98, opened 2023-06-26
Expand on 'required' and 'breaking' additions in versioning

palday:
Based on a discussion I had with @jrevels -- this resolves a point of confusion I had with the addition of an 'optional' field, i.e. a field `<:Union{Missing, ...}`. The example here and the clarified narrower meaning of "required" cleared up my misunderstanding.

### DIFF
--- a/docs/src/schema-concepts.md
+++ b/docs/src/schema-concepts.md
@@ -29,13 +29,15 @@ While it is fairly established practice to [semantically version source code](ht
 
 For example, a schema author must introduce a new schema version for any of the following changes:
 
-- A new type-restricted required field is added to the schema.
+- A new type-restricted required field is added to the schema. *Required* means any field which is specified in the schema declaration, even if it is allowed to be `Missing`. In other words, the addition of any field with a type-restriciton narrow than `Any` is breaking.
 - An existing required field's type restriction is tightened.
 - An existing required field is renamed.
 
 One benefit of Legolas' approach is that multiple schema versions may be defined in the same codebase, e.g. there's nothing that prevents `@version(FooV1, ...)` and `@version(FooV2, ...)` from being defined and utilized simultaneously. The source code that defines any given Legolas schema version and/or consumes/produces Legolas tables is presumably already semantically versioned, such that consumer/producer packages can determine their compatibility with each other in the usual manner via interpreting major/minor/patch increments.
 
 Note that it is preferable to avoid introducing new versions of an existing schema, if possible, in order to minimize code/data churn for downstream producers/consumers. Thus, authors should prefer conservative field type restrictions from the get-go. Remember: loosening a field type restriction is not a breaking change, but tightening one is.
+
+Introducing a new field with a restriction narrower than `Any` is in some sense a tightening of type restrictions and thus breaking because extensions may have already defined that field with an incompatible restriction. 
 
 ## Important Expectations Regarding Custom Field Assignments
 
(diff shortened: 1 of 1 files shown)


ericphanson:
Oh, I don't think I realized this changed [since v0.4](https://beacon-biosignals.github.io/Legolas.jl/v0.4/schema/#Simple-Integer-Versioning:-You-Break-It,-You-Bump-It-1). That's a big change!

kleinschmidt:
I guess I was confused about this too, but now having thought about it a bit more, I don't think there's any way to have _both_ schema inheritance _and_ schema versioning without this interpretation.  we've just hit a case where an upstream schema added an "optional field" (`Union{T,Missing}`) which broke my inheriting schema that has a field with the same name but different type/meaning.

jrevels:
>  we've just hit a case where an upstream schema added an "optional field" (Union{T,Missing}) which broke my inheriting schema that has a field with the same name but different type/meaning.

yup, this is exactly why Legolas treats these a breaking. The only way you can add a new required field in a non-breaking way is if it is entirely unprocessed/unconstrained 

it's worth adding an explicit example to the `tour.jl` in this PR; I should be able to do this quickly tomorrow

---

I think the root of the misunderstanding here is that it sounds like multiple users didn't realize/know Legolas' definition of `required field`; I think folks saw that term and reasonably assumed "ah, a field is not required if it's `Missing`" - which is definitely sensible in common parlance but is strictly incorrect in the Legolas sense

this PR helps mitigate this confusion by making the definition loud/clear and appropriately positioned (as opposed to requiring the reader to read/remember the tour), so we should def fold it in ASAP (i'll do so tomorrow)

long term, though, I wonder if there's a better term to use that is less likely to be confused in this way. "schema-version-specified"? idk. will think on it.

ericphanson:
> long term, though, I wonder if there's a better term to use that is less likely to be confused in this way. "schema-version-specified"? idk. will think on it.

I think "required field" is extra-confusing in Legolas v0.5 land, because there are _only_ required fields in the structs. "Non-required" fields only exist outside of the types that Legolas itself provides. So if one is thinking of those as the entirety of the fields, then the modifier `required` seems to be something stricter than that, for which not-allowed-to-be-missing feels natural (and kind of aligns with how in Legolas pre-v0.5, allowed-missing fields weren't "required" to be present in a table in any form in order to construct the row).

To me just calling it a "field" seems clear, without adding "required". E.g.

> If you are adding a field to a schema, it must be entirely untyped and unconstrained to not be breaking.

It doesn't feel like any additional qualifier is necessary in that context.

jrevels:
After mulling it over, I'm pretty convinced we should just call these "declared fields" instead

Rebranding from "required" to "declared" can easily be achieved without breaking any downstream code (just requires putting some aliases in place), and seems like it'll nip this confusion in the bud moving forward, so I'm going to hijack this PR and go for it

---

historical tidbit: the original motivation behind the term "required field" is that a given record is *required* to contain a matching field for every required field of a given schema version in order to comply with that schema version (with the additional practical convenience that most Legolas tooling interprets non-present required fields as present-but-missing). i.e. a required field == a field required to comply with schema version.

it's definitely still useful to be able to talk about a record's fields as required vs. non-required w.r.t. a given schema (or when comparing two schemas, or schema extensions, etc.). but we can use "declared" to refer to "fields declared to be required by a specific schema version" - this has less confusion potential for all the reasons already discussed

jrevels:
superseded by https://github.com/beacon-biosignals/Legolas.jl/pull/100

PR #99, opened 2023-07-06
ConstructionBase integration

### LINKED ISSUE #97: Accessors.jl support (lack of positional argument constructors)
The optics-based Accessors.jl makes it easy to "set" fields of immutable structs, and could replace a lot of `Tables.rowmerge` usage for legolas-defined record types.  But because teh record types don't have positional argument constructors defined, `@set` and friends don't work:

```julia
julia> using Accessors

julia> @schema "foo" Foo

julia> @version FooV1 begin
           a::String
           b::Int
       end
F
julia> FooV1(; a="a", b=1)
FooV1: (a = "a", b = 1)

julia> foo = FooV1(; a="a", b=1)
FooV1: (a = "a", b = 1)

julia> @set foo.a="aa"
ERROR: MethodError: no method matching FooV1(::String, ::Int64)

Closest candidates are:
  FooV1(::Any)
   @ Main ~/.julia/packages/Legolas/nhYG7/src/schemas.jl:605

Stacktrace:
 [1] setproperties_object(obj::FooV1, patch::NamedTuple{(:a,), Tuple{String}})
   @ ConstructionBase ~/.julia/packages/ConstructionBase/pAlst/src/ConstructionBase.jl:212
 [2] setproperties(obj::FooV1, patch::NamedTuple{(:a,), Tuple{String}})
   @ ConstructionBase ~/.julia/packages/ConstructionBase/pAlst/src/ConstructionBase.jl:126
 [3] set(obj::FooV1, l::PropertyLens{:a}, val::String)
   @ Accessors ~/.julia/packages/Accessors/4DnIb/src/optics.jl:375
 [4] top-level scope
   @ REPL[8]:1
```

We could fix this by defining a positional argumetn constructor (namedtuples have them!), or appropriate methods for `setproperties_object(::AbstractRecord, ...)`.

kleinschmidt:
Due to the possibility of parametric record types, this is a bit more complex than just creating a namedtuple, so I opted to include a method for each `@version` inside the outer constructors block.

There's an unfortunate bit of copy-paste required from ConstructionBase due to how ConstructionBase checks for propertynames vs. fieldnames in versions before 1.7; that can be removed if we stop supporting 1.6.

closes #97 

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1205001403646390


kleinschmidt:
Hm, not clear why 1.6 is failing.  Here's the error msg:

````
  The function `Base.propertynames` was overloaded for type `ParentV1`.
  Please make sure the following methods are also overloaded for this type:
  ```julia
  ConstructionBase.setproperties
  ConstructionBase.getproperties # optional in VERSION >= julia v1.7
  ```
````

File: src/schemas.jl [hunk 1]
```diff
@@ -344,6 +344,23 @@ abstract type AbstractRecord <: Tables.AbstractRow end
...
 @inline Tables.schema(::AbstractVector{R}) where {R<:AbstractRecord} = Tables.Schema(fieldnames(R), fieldtypes(R))
 
+# we need a bit of extra work to integrate with ConstructionBase for pre-1.7 due
+# to the overload of `propertynames(::Tables.AbstractRow)`
+#
+# we _could_ overload `check_properties_are_fields` but that's no part of the
+# public API so this is safer
+@static if VERSION < v"1.7"
+    ConstructionBase.getproperties(r::AbstractRecord) = NamedTuple(r)
+    # largely copy-paste from ConstructionBase.setproperties_object:
```
omus, line 354: Can you provide permalinks to the code this is based off of? Makes this easier to maintain

test/runtests.jl
```diff
@@ -686,3 +699,30 @@ end
...
+##### ConstructionBase/Accessors integration
+#####
+
+@testset "ConstructionBase/Accessors integration" begin
+    p = ParentV1(; x=[1,2,3], y="hello")
+    p2 = @set p.y = "okay"
+    @test p2 isa ParentV1
+    @test p2.y == "okay"
+    @test p2 !== p
+    @test p2.x === p.x
```
omus, line 713:
```suggestion
    @test p2.x === p.x
    @test p2.y != p.y
```

test/runtests.jl
```diff
@@ -686,3 +699,30 @@ end
...
+
+    e = ArgumentError("Invalid value set for field `y`, expected AbstractString, got a value of type Int64 (1)")
+    @test_throws e @set p.y = 1
+
+    n = NestedV1(; gc=GrandchildV1(; x=[1,2], y="hello", z=nothing, a=1), k=:okay)
+    n2 = @set n.gc.z = "yay!"
+    @test n2.gc.z == "yay!"
+    @test n2.gc !== n.gc
```
omus, line 721:
```suggestion
    @test n2.gc != n.gc
```

test/runtests.jl
```diff
@@ -686,3 +699,30 @@ end
...
+    # parametric on :k
+    @test typeof(n) == typeof(n2)
```
omus, line 723:
```suggestion
    @test typeof(n2) == typeof(n)
    @test n2.k === n.k
```

test/runtests.jl
```diff
@@ -203,6 +212,10 @@ module Namespace91
         @test A91V1(; a=1) == A91V1(; a=1)
         @test isequal(A91V1(; a=1), A91V1(; a=1))
         @test read(tobuffer([A91V1(; a=1)], A91V1SchemaVersion())).a == [1]
+
+        a = A91V1(; a=1)
+        a2 = @set a.a = 2
+        @test a2.a == 2
```
omus, line 218: Why are these tests needed? Seems like the later testset covers this. At the very least this should be made into a separate testset inside this module

src/Legolas.jl
```diff
@@ -1,6 +1,6 @@
 module Legolas
 
-using Tables, Arrow, UUIDs
+using Tables, Arrow, UUIDs, ConstructionBase
```
omus, line 3: I'm very much for being explicit about imports in libraries:
```suggestion
using Tables, Arrow, UUIDs
using ConstructionBase: ConstructionBase
```

src/schemas.jl
```diff
@@ -344,6 +344,23 @@ abstract type AbstractRecord <: Tables.AbstractRow end
...
+    function ConstructionBase.setproperties(r::R, patch::NamedTuple) where {R <: AbstractRecord}
+        nt = ConstructionBase.getproperties(r)
+        nt_new = merge(nt, patch)
+        ConstructionBase.check_patch_properties_exist(nt_new, nt, r, patch)
+        args = Tuple(nt_new) # old julia inference prefers if we wrap in Tuple
+        return ConstructionBase.constructorof(R)(args...)
```
omus, line 360: At first I was curious about improving performance as this converts to a `NamedTuple` first. As this is also what [`Tables.rowmerge`](https://github.com/JuliaData/Tables.jl/blob/dac28d9f233c297ab8d22ce9cab439ac3e97899d/src/utils.jl#L157C30-L157C49) already does this seems perfectly fine.

omus (approved):

test/runtests.jl
kleinschmidt: just to make sure that the ConstructionBase name is interpolated correctly as well, since we generate code with `$ConstructionBase.` in teh macro

src/schemas.jl
kleinschmidt: yeah, I was starting to go down that rabbit hole and eventually decided this was probably good enough...

Project.toml
```diff
@@ -5,21 +5,25 @@ version = "0.5.13"
...
 Arrow = "2"
 Compat = "3.34, 4"
+ConstructionBase = "1.5"
 DataFrames = "1"
 Tables = "1.4"
 julia = "1.6"
 
 [extras]
+Accessors = "7d9f7c33-5ae7-4f3b-8dc6-eff91059b697"
+Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
```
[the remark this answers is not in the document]
kleinschmidt: Hm, I'm not sure.  Accessors in particular is still on 0.1, and I'm wary of adding additional compat hell just for tests...on the flip side, we wouldn't want our tests to break just because Accessors made a breaking release so...

src/Legolas.jl
kleinschmidt: makes sense, especially since we're only defining methods, not calling anything AFAICT

Project.toml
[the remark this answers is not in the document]
omus: Probably mostly beneficial to have the compat bounds. If we end up upper-bounding the required test packages we avoid accidentally breaking tests and additionally continue testing an older version retains compatibility with the rest of our dependencies

test/runtests.jl
omus: Okay, would have been nice to have this as a separate testcase inside the submodule as mentioned. I'll leave this unresolved

PR #100, opened 2023-07-08
replace the usage of the term "required field" with "declared field" and clarify breaking changes for schema versions

jrevels:
Supersedes #98 ; xref https://github.com/beacon-biosignals/Legolas.jl/pull/98#issuecomment-1627373261


palday:
Instead of doing a breaking change, why not just `@deprecate` the old name?

jrevels:
> Instead of doing a breaking change, why not just @deprecate the old name?

That's what this PR does?

or do you just mean the version bump? usually i view introducing a new depwarn as a minor version bump in a `0.x` series, but I guess i could bump the patch version instead if that's what folks want

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Legolas"
 uuid = "741b9549-f6ed-4911-9fbf-4a1c0c97f0cd"
 authors = ["Beacon Biosignals, Inc."]
-version = "0.5.12"
+version = "0.6.0"
```
jrevels, line 4:
```suggestion
version = "0.5.13"
```

jrevels:
sounds good, changed the Project.toml bump to `0.5.13` instead of `0.6.0`

src/schemas.jl
```diff
@@ -409,24 +411,27 @@ function Base.showerror(io::IO, e::SchemaVersionDeclarationError)
...
+struct DeclaredFieldInfo
     name::Symbol
     type::Union{Symbol,Expr}
     parameterize::Bool
     statement::Expr
 end
 
-Base.:(==)(a::RequiredFieldInfo, b::RequiredFieldInfo) = all(getfield(a, i) == getfield(b, i) for i in 1:fieldcount(RequiredFieldInfo))
+# We maintain an alias to the deprecated name for this type, xref https://github.com/beacon-biosignals/Legolas.jl/pull/100
+const RequiredFieldInfo = DeclaredFieldInfo
```
palday, line 430:
```suggestion
Base.@deprecate_binding RequiredFieldInfo DeclaredFieldInfo
```

PR #102, opened 2023-09-28
Allow a schema version to declare no fields

ararslan:
The motivating use case for this was a set of schemas used as an enumeration of sorts, all of which inherit from a common parent but do not differ structurally from the parent.

### DIFF
--- a/src/schemas.jl
+++ b/src/schemas.jl
@@ -777,7 +777,7 @@ Note that this macro expects to be evaluated within top-level scope.
 For more details and examples, please see `Legolas.jl/examples/tour.jl` and the
 "Schema-Related Concepts/Conventions" section of the Legolas.jl documentation.
 """
-macro version(record_type, declared_fields_block)
+macro version(record_type, declared_fields_block=nothing)
     # parse `record_type`
     if record_type isa Symbol
         parent_record_type = nothing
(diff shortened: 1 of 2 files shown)


src/schemas.jl
```diff
@@ -661,8 +661,8 @@ function _generate_record_type_definitions(schema_version::SchemaVersion, record
     end
 
     # generate `base_overload_definitions`
-    equal_rhs_statement = foldr((x, y) -> :($x && $y), (:(a.$f == b.$f) for f in keys(record_fields)))
-    isequal_rhs_statement = foldr((x, y) -> :($x && $y), (:(isequal(a.$f, b.$f)) for f in keys(record_fields)))
+    equal_rhs_statement = foldr((x, y) -> :($x && $y), (:(a.$f == b.$f) for f in keys(record_fields)); init=true)
+    isequal_rhs_statement = foldr((x, y) -> :($x && $y), (:(isequal(a.$f, b.$f)) for f in keys(record_fields)); init=true)
```
ararslan, line 665: Note that the case that this change is intended to fix is already handled by the approach in #103.

jrevels (commented):
LGTM - include release version bump in same PR?

ararslan:
The release bump happened in #103 but no release has been made from that yet.

PR #103, opened 2023-09-29
Use common methods for Base functions instead of generating

### LINKED ISSUE #101: `==` semantics don't match NamedTuple's in presence of missing field
```julia
julia> using Legolas

julia> @schema "example.foo" Foo

julia> @version FooV1 begin
            a::Union{Missing, Int}
            b::Union{Missing, Int}
        end

julia> x = (; a = missing, b = 1)
(a = missing, b = 1)

julia> y = (; a = missing, b = 2)
(a = missing, b = 2)

julia> x == y
false

julia> FooV1(x) == FooV1(y)
ERROR: TypeError: non-boolean (Missing) used in boolean context
Stacktrace:
 [1] ==(a::FooV1, b::FooV1)
   @ Main ~/.julia/packages/Legolas/0cASP/src/schemas.jl:668
 [2] top-level scope
   @ REPL[11]:1
 [3] top-level scope
   @ ~/.julia/juliaup/julia-1.9.2+0.aarch64.apple.darwin14/share/julia/stdlib/v1.9/REPL/src/REPL.jl:1416
```
due to the implementation in https://github.com/beacon-biosignals/Legolas.jl/blob/0d9ec276241dfe1dc46aae64cad4e3d5394c7fb9/src/schemas.jl#L664. We should use something like https://github.com/JuliaServices/AutoHashEquals.jl/pull/45/files#diff-338e197a5b351ab9a74086bcc5d606d7c05b969f9d4b4b5bcf67b8f7083ec852R321-R355 which matches Base's https://github.com/JuliaLang/julia/blob/96e5b62e95f79a64ab296001dd8cf748c63af806/base/tuple.jl#L482-L514.

ararslan:
Currently, methods for `==`, `isequal`, `hash`, and `NamedTuple` are generated by the `@version` macro for every row type that gets defined. The bodies of these methods always follow the same patterns, none of which require any information that's only available in the context where the definitions are generated. In fact, specific methods don't need to be generated at all; `AbstractRecord`s can all use the same generic methods for these functions. This provides the following additional benefits:
- When many schema versions are defined, this significantly reduces the excessive number of redundant methods which make method autocompletion effectively useless.
- Defining the generic methods in terms of regular functions rather than generating expressions in the macro means that it's easier to reason about the methods' behavior and ensure overall consistency. For example, moving from generating a chain of `==` and `&&` over a record's fields to directly using `all` means that `missing` is treated consistently (see #101).

Note that the generic method for `hash` as defined here loops over the fields in reverse order. This is to match the previous `foldr` behavior, ensuring that hashes don't change with this implementation.

Fixes #101.

### DIFF
--- a/src/schemas.jl
+++ b/src/schemas.jl
@@ -856,12 +856,19 @@ end
 ##### Base overload definitions
 #####
 
-function Base.:(==)(x::T, y::T) where {T<:AbstractRecord}
-    return all(i -> getfield(x, i) == getfield(y, i), 1:fieldcount(T))
+_typeof(r::AbstractRecord) = Base.unwrap_unionall(Base.typename(typeof(r)).wrapper)
+
+_type_equal(x::R, y::R) where {R<:AbstractRecord} = true
+_type_equal(x::AbstractRecord, y::AbstractRecord) = _typeof(x) === _typeof(y)
+
+function Base.:(==)(x::AbstractRecord, y::AbstractRecord)
+    _type_equal(x, y) || return false
+    return all(i -> getfield(x, i) == getfield(y, i), 1:nfields(x))
 end
 
-function Base.isequal(x::T, y::T) where {T<:AbstractRecord}
-    return all(i -> isequal(getfield(x, i), getfield(y, i)), 1:fieldcount(T))
+function Base.isequal(x::AbstractRecord, y::AbstractRecord)
+    _type_equal(x, y) || return false
+    return all(i -> isequal(getfield(x, i), getfield(y, i)), 1:nfields(x))
 end
 
 function Base.hash(r::AbstractRecord, h::UInt)
--- a/src/schemas.jl
+++ b/src/schemas.jl
@@ -856,7 +856,7 @@ end
 ##### Base overload definitions
 #####
 
-_typeof(r::AbstractRecord) = Base.unwrap_unionall(Base.typename(typeof(r)).wrapper)
+_typeof(r::AbstractRecord) = record_type(schema_version_from_record(r))
 
 _type_equal(x::R, y::R) where {R<:AbstractRecord} = true
 _type_equal(x::AbstractRecord, y::AbstractRecord) = _typeof(x) === _typeof(y)
--- a/src/schemas.jl
+++ b/src/schemas.jl
@@ -447,7 +447,7 @@ end
 # We maintain an alias to the deprecated name for this type, xref https://github.com/beacon-biosignals/Legolas.jl/pull/100
 Base.@deprecate_binding RequiredFieldInfo DeclaredFieldInfo
 
-Base.:(==)(a::DeclaredFieldInfo, b::DeclaredFieldInfo) = all(getfield(a, i) == getfield(b, i) for i in 1:fieldcount(DeclaredFieldInfo))
+Base.:(==)(a::DeclaredFieldInfo, b::DeclaredFieldInfo) = _compare_fields(==, a, b)
 
 function _parse_declared_field_info!(f)
     f isa Symbol && (f = Expr(:(::), f, :Any))
@@ -856,20 +856,24 @@ end
 ##### Base overload definitions
 #####
 
-_typeof(r::AbstractRecord) = record_type(schema_version_from_record(r))
-
-_type_equal(x::R, y::R) where {R<:AbstractRecord} = true
-_type_equal(x::AbstractRecord, y::AbstractRecord) = _typeof(x) === _typeof(y)
-
-function Base.:(==)(x::AbstractRecord, y::AbstractRecord)
-    _type_equal(x, y) || return false
-    return all(i -> getfield(x, i) == getfield(y, i), 1:nfields(x))
+# Field-wise comparison for any two objects with exactly the same type. Most record
+# comparisons will hit this method.
+function _compare_fields(eq, x::T, y::T) where {T}
+    return all(i -> eq(getfield(x, i), getfield(y, i)), 1:fieldcount(T))
 end
 
-function Base.isequal(x::AbstractRecord, y::AbstractRecord)
-    _type_equal(x, y) || return false
-    return all(i -> isequal(getfield(x, i), getfield(y, i)), 1:nfields(x))
+# Field-wise comparison of two arbitrary records, with equality contingent on matching
+# schemas. Record comparisons for parametric record types with mismatched type parameters
+# will hit this method, as well as mismatched record types (which will not compare equal).
+function _compare_fields(eq, x::AbstractRecord, y::AbstractRecord)
+    svx = schema_version_from_record(x)
+    svy = schema_version_from_record(y)
+    return svx === svy && all(i -> eq(getfield(x, i), getfield(y, i)), 1:nfields(x))
 end
+
+Base.:(==)(x::AbstractRecord, y::AbstractRecord) = _compare_fields(==, x, y)
+
+Base.isequal(x::AbstractRecord, y::AbstractRecord) = _compare_fields(isequal, x, y)
 
 function Base.hash(r::AbstractRecord, h::UInt)
     for i in nfields(r):-1:1
--- a/src/schemas.jl
+++ b/src/schemas.jl
@@ -660,17 +660,6 @@ function _generate_record_type_definitions(schema_version::SchemaVersion, record
         end
     end
 
-    # generate `base_overload_definitions`
-    equal_rhs_statement = foldr((x, y) -> :($x && $y), (:(a.$f == b.$f) for f in keys(record_fields)))
-    isequal_rhs_statement = foldr((x, y) -> :($x && $y), (:(isequal(a.$f, b.$f)) for f in keys(record_fields)))
-    hash_rhs_statement = foldr((x, y) -> :(hash($x, $y)), (:(r.$f) for f in keys(record_fields)); init=:h)
-    base_overload_definitions = quote
-        Base.:(==)(a::$R, b::$R) = $equal_rhs_statement
-        Base.isequal(a::$R, b::$R) = $isequal_rhs_statement
-        Base.hash(r::$R, h::UInt) = hash($R, $hash_rhs_statement)
-        Base.NamedTuple(r::$R) = (; $((:(r.$f) for f in keys(record_fields))...))
-    end
-
     # generate `arrow_overload_definitions`
     record_type_arrow_name = string("JuliaLang.Legolas.Generated.", Legolas.name(schema_version), '.', Legolas.version(schema_version))
     record_type_arrow_name = Base.Meta.quot(Symbol(record_type_arrow_name))
@@ -689,7 +678,6 @@ function _generate_record_type_definitions(schema_version::SchemaVersion, record
             $inner_constructor_definitions
         end
         $outer_constructor_definitions
-        $base_overload_definitions
         $arrow_overload_definitions
         $Legolas.record_type(::$(Base.Meta.quot(typeof(schema_version)))) = $R
         $Legolas.schema_version_from_record(::$R) = $schema_version
@@ -862,4 +850,28 @@ macro version(record_type, declared_fields_block)
         end
         nothing
     end
+end
+
+#####
+##### Base overload definitions
+#####
+
+function Base.:(==)(x::T, y::T) where {T<:AbstractRecord}
+    return all(i -> getfield(x, i) == getfield(y, i), 1:fieldcount(T))
+end
+
+function Base.isequal(x::T, y::T) where {T<:AbstractRecord}
+    return all(i -> isequal(getfield(x, i), getfield(y, i)), 1:fieldcount(T))
+end
+
+function Base.hash(r::AbstractRecord, h::UInt)
+    for i in nfields(r):-1:1
+        h = hash(getfield(r, i), h)
+    end
+    return hash(typeof(r), h)
+end
+
+function Base.NamedTuple(r::AbstractRecord)
+    names = fieldnames(typeof(r))
+    return NamedTuple{names}(map(x -> getfield(r, x), names))
 end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -516,9 +516,7 @@ end
         c = ChildV1(; x=[1, 2], y="hello", z=missing)
         @test isequal(c, c)
         @test ismissing(c == c)
-        if UInt === UInt64  # value will be different depending on system word size
-            @test hash(c) === 0x07055951b3aa478e
-        end
+        @test hash(c) isa UInt  # NOTE: can't rely on particular values
     end
 end
 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -526,10 +526,18 @@ end
         @test isequal(a, b)
         @test ismissing(a == b)
         @test hash(a) == hash(b)
-        @test UnionMissingV1(; a=missing, b=1) != UnionMissingV1(; a=missing, b=2)
-        @test !isequal(UnionMissingV1(; a=missing, b=1), UnionMissingV1(; a=missing, b=2))
-        @test ParamV1(; i=one(Int32)) == ParamV1(; i=one(Int64))
-        @test isequal(ParamV1(; i=one(Int32)), ParamV1(; i=one(Int64)))
+        u1 = UnionMissingV1(; a=missing, b=1)
+        u2 = UnionMissingV1(; a=missing, b=2)
+        @test u1 != u2
+        @test !isequal(u1, u2)
+        p32 = ParamV1(; i=one(Int32))
+        p64 = ParamV1(; i=one(Int64))
+        @test p32 == p64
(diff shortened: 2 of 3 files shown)


ericphanson:
Since this closes #101, can you add the test from that post? (The current one doesn't cover the case in which there is `missing` present but `==` should return false since there is a non-missing field that differs)

src/schemas.jl
```diff
@@ -863,3 +851,27 @@ macro version(record_type, declared_fields_block)
         nothing
     end
 end
+
+#####
+##### Base overload definitions
+#####
+
+function Base.:(==)(x::T, y::T) where {T<:AbstractRecord}
```
ararslan, line 859: Actually this isn't quite right: for parametric row types, this will method won't be hit if the type parameters differ. Using this setup (taken from the tests):
```julia
@schema "test.param" Param
@version ParamV1 begin
    i::(<:Integer)
end
```
we get
```julia
julia> ParamV1(; i=one(Int32)) == ParamV1(; i=one(Int64))
false
```
on this branch but `true` on main.

src/schemas.jl
```diff
@@ -856,12 +856,19 @@ end
 ##### Base overload definitions
 #####
 
-function Base.:(==)(x::T, y::T) where {T<:AbstractRecord}
-    return all(i -> getfield(x, i) == getfield(y, i), 1:fieldcount(T))
+_typeof(r::AbstractRecord) = Base.unwrap_unionall(Base.typename(typeof(r)).wrapper)
```
ericphanson, line 859: What if we generated methods for `_typeof` in the macro? Could avoid internals while still having fewer methods overall

jrevels:
can we check that performance is the same as the old methods in benchmark contexts where record types aren't fully inferred

ararslan:
Not sure if this is what you had in mind but here's what I get. Looks fine to me.

Setup:
```julia
using Legolas, BenchmarkTools
using Legolas: @schema, @version

@schema "test.param" Param
@version ParamV1 begin
    i::(<:Integer)
end
@version ParamV2 begin
    i::(<:Integer)
end

bad1() = ParamV1(; i=(rand() < 0.5 ? Int32 : Int64)(1))
bad2() = (rand() < 0.5 ? ParamV1 : ParamV2)(; i=(rand() < 0.5 ? Int32 : Int64))
```
The return type of `bad1` infers as `Union{ParamV1{Int32},ParamV1{Int64}}` and the return type of `bad2` can't be inferred.

main:
```julia
julia> @benchmark bad1() == bad1()
BenchmarkTools.Trial: 10000 samples with 10 evaluations.
 Range (min … max):  870.300 ns … 37.980 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):       1.292 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):     1.913 μs ±  2.597 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ██▆▆▅▃▂▂▂▂▂▁                                                 ▂
  ███████████████▇▆▅▅▁▄▄▄▄▄▄▄▄▅▅▄▅▄▁▅▁▃▃▁▃▄▁▃▄▁▁▄▁▁▄▃▃▁▃▅▅▃▅▅▅ █
  870 ns        Histogram: log(frequency) by time      19.5 μs <

 Memory estimate: 64 bytes, allocs estimate: 4.

julia> @benchmark bad2() == bad2()
BenchmarkTools.Trial: 10000 samples with 10 evaluations.
 Range (min … max):  1.441 μs …  14.840 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     1.660 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.668 μs ± 224.386 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                         ▁▂▂▅▅▆▇▇▇█▇█▇█▆▄▂▃▁▁
  ▁▁▁▁▁▁▁▁▁▂▂▂▂▂▃▃▄▄▅▅▆▆██████████████████████▆▆▅▄▄▃▃▂▂▂▂▂▁▁▁ ▄
  1.44 μs         Histogram: frequency by time        1.83 μs <

 Memory estimate: 96 bytes, allocs estimate: 6.
```

this branch:
```julia
julia> @benchmark bad1() == bad1()
BenchmarkTools.Trial: 10000 samples with 10 evaluations.
 Range (min … max):  1.269 μs …  11.213 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     1.493 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.510 μs ± 210.917 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

          ▂▄▆█▇▆▅▂
  ▂▂▂▃▃▅▆▇████████▇▆▄▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▁▂▁▁▁▁▁▁▂▂▂▂▂▂▁▂▂▂ ▃
  1.27 μs         Histogram: frequency by time        2.37 μs <

 Memory estimate: 64 bytes, allocs estimate: 4.

julia> @benchmark bad2() == bad2()
BenchmarkTools.Trial: 10000 samples with 10 evaluations.
 Range (min … max):  963.900 ns … 36.246 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):       1.435 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):     1.937 μs ±  2.029 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▅█▇▅▅▅▄▃▂▁ ▁▁▁ ▁                                             ▂
  ██████████████████▆█▆▆▆▆▅▄▅▁▁▃▄▄▄▅▄▅▆▅▆▅▆▅▅▃▄▄▄▃▄▅▄▄▅▅▄▄▄▄▃▄ █
  964 ns        Histogram: log(frequency) by time      13.8 μs <

 Memory estimate: 96 bytes, allocs estimate: 6.
```

ararslan:
More directly targeting the equality check, there's unsurprisingly no difference at all.

main:
```julia
julia> @benchmark x == y setup=(x = bad1(); y = bad1())
BenchmarkTools.Trial: 10000 samples with 1000 evaluations.
 Range (min … max):  6.113 ns … 366.950 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     6.207 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   8.007 ns ±   8.376 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █  ▆           ▁▅▁   ▃                                      ▁
  █▅▇██▆▆▃▆▅▄▇▇▆▆███▆▅▄██▅▄▄▅▆▅▆▆▆▆▆▆████▇▇█▇▇▆▇▆▆▅▆▅▄▄▄▅▆▇▇▆ █
  6.11 ns      Histogram: log(frequency) by time      17.5 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark x == y setup=(x = bad2(); y = bad2())
BenchmarkTools.Trial: 10000 samples with 995 evaluations.
 Range (min … max):  26.546 ns … 97.779 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     34.875 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   35.495 ns ±  4.587 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

               █▃       ▁▆▃
  ▂▃▃▄▃▂▁▁▂▂▂▃▆██▆▅▄▃▃▂▄████▇▅▆▅▆▅▅▅▄▄▄▅▅▆▅▄▄▃▃▃▃▃▃▂▂▂▂▂▂▂▂▂▁ ▃
  26.5 ns         Histogram: frequency by time        46.1 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

this branch:
```julia
julia> @benchmark x == y setup=(x = bad1(); y = bad1())
BenchmarkTools.Trial: 10000 samples with 1000 evaluations.
 Range (min … max):  6.113 ns … 192.035 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     6.820 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   7.461 ns ±   4.978 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █  █▆▁▁     ▁▁▂▂▂▃▂▂▁ ▁     ▁▁▁                             ▂
  █████████████████████████▇▇██████▇▇▇▆▆▆▆▆▄▆▁▄▆▅▆▄▄▇▄▄▄▄▃▅▅▅ █
  6.11 ns      Histogram: log(frequency) by time      16.8 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark x == y setup=(x = bad2(); y = bad2())
BenchmarkTools.Trial: 10000 samples with 995 evaluations.
 Range (min … max):  26.282 ns … 201.913 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     34.388 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   35.156 ns ±   4.858 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                █▅      ▁▅█▂    ▁
  ▂▃▅▃▃▁▁▁▂▄▅▅▄▆██▅▄▅▃▃▅████▇▅▄▅█▇▅▅▆▅▄▅▆▇▆▅▄▄▃▄▃▄▄▃▃▂▂▂▂▂▂▂▂▁ ▃
  26.3 ns         Histogram: frequency by time         45.6 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

test/runtests.jl
```diff
@@ -511,6 +518,17 @@ end
         @test_throws TypeError ParamV1(; i=1.0)
         @test_throws ArgumentError ParamV1{Int}(; i=1.1)
     end
+
+    @testset "equality and hashing" begin
+        c = ChildV1(; x=[1, 2], y="hello", z=missing)
+        @test isequal(c, c)
+        @test ismissing(c == c)
+        @test hash(c) isa UInt  # NOTE: can't rely on particular values
```
omus, line 526:
```suggestion
        a = ChildV1(; x=[1, 2], y="hello", z=missing)
        b = ChildV1(; x=[1, 2], y="hello", z=missing)
        @test a !== b
        @test isequal(a, b)
        @test ismissing(a == b)
        @test hash(a) == hash(b)
```

test/runtests.jl
```diff
@@ -511,6 +518,17 @@ end
...
+        @test UnionMissingV1(; a=missing, b=1) != UnionMissingV1(; a=missing, b=2)
+        @test !isequal(UnionMissingV1(; a=missing, b=1), UnionMissingV1(; a=missing, b=2))
+        @test ParamV1(; i=one(Int32)) == ParamV1(; i=one(Int64))
+        @test isequal(ParamV1(; i=one(Int32)), ParamV1(; i=one(Int64)))
+    end
```
omus, line 531: Should add some tests comparing `ChildV1` to a `ParentV1` and a `ParentV1` to a `ChildV1`. In particuar I am thinking about `_compare_fields(eq, ::AbstractRecord, ::AbstractRecord)`

ararslan: Is what I just added what you had in mind?

jrevels:
I sanity checked NamedTuple conversion performance as it can end up being in the critical inner loops in some use cases; result is that it seems fine on julia 1.9.

```jl
using Legolas, BenchmarkTools
using Legolas: @schema, @version

@schema "test.foo" Foo

@version FooV1 begin
    a::(<:Any)
    b::(<:Any)
    c::(<:Any)
    d::(<:Any)
end

@version FooV2 begin
    x::(<:Any)
    y::(<:Any)
    z::(<:Any)
end

# attempt to thwart compiler optimizations
function test(v, n)
    result = 0
    for _ in 1:n
        result += sum(sizeof, values(NamedTuple(rand(v))))
    end
    return result
end

gen_foos() = Any[FooV1(a=1,b=2,c=3,d=4),FooV2(x="x",y="y",z="z")]

@benchmark test(foos, 100) setup=(foos=gen_foos()) evals=1
```

this branch:

```julia
julia> @benchmark test(foos, 100) setup=(foos=gen_foos()) evals=1
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  15.541 μs …  1.652 ms  ┊ GC (min … max): 0.00% … 97.53%
 Time  (median):     16.167 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   16.839 μs ± 28.094 μs  ┊ GC (mean ± σ):  2.84% ±  1.69%

        ▄ █▁▆▂
  ▁▁▁▂▃▆█▇████▄▅▂▃▃▂▃▂▃▂▁▂▁▁▁▁▂▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  15.5 μs         Histogram: frequency by time        19.5 μs <

 Memory estimate: 15.11 KiB, allocs estimate: 441.
```

main:

```jl
julia> @benchmark test(foos, 100) setup=(foos=gen_foos()) evals=1
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  15.750 μs …  1.811 ms  ┊ GC (min … max): 0.00% … 96.41%
 Time  (median):     16.625 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   17.324 μs ± 30.596 μs  ┊ GC (mean ± σ):  2.99% ±  1.68%

          ▄ ██▅
  ▂▂▂▂▂▃▅██████▆█▅▄▄▃▄▄▃▂▂▂▂▂▂▃▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▁▂▂▂▂▂▂▂▂▂▂ ▃
  15.8 μs         Histogram: frequency by time        20.1 μs <

 Memory estimate: 14.98 KiB, allocs estimate: 443.
```

ericphanson:
FWIW if you are looking for perf drops here, I would look at > 10 fields. E.g. in https://discourse.julialang.org/t/slowness-of-fieldnames-and-propertynames/55364/2 (very similar case) I found inference failed for 11 fields but not for 9. That was on an older version of julia though.

ararslan:
Tried using row types with 15 and 25 parameters and still get no meaningful difference.

setup
```julia
using Legolas, BenchmarkTools, Base.Iterators
using Legolas: @schema, @version

macro make_version(name, n)
    block = Expr(:block)
    ex = Expr(:macrocall, Symbol("@version"), __source__, name, block)
    for i in 1:n
        x = Symbol(:x, i)
        push!(block.args, Expr(:(::), Symbol(:x, i), :(<:Any)))
    end
    return ex
end

@schema "test.foo" Foo
@make_version FooV1 15
@make_version FooV2 25

function make_row(T, values)
    nt = NamedTuple()
    for (field, value) in zip(fieldnames(T), values)
        nt = merge(nt, (; field => value))
    end
    return T(; nt...)
end

gen_foos() = Any[make_row(FooV1, countfrom(1)), make_row(FooV2, string.('a':'z'))]

function test(v, n)
    result = 0
    for _ in 1:n
        result += sum(sizeof, values(NamedTuple(rand(v))))
    end
    return result
end

@benchmark test(foos, 100) setup=(foos = gen_foos()) evals=1
```

this branch
```
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  17.958 μs … 887.416 μs  ┊ GC (min … max): 0.00% … 95.49%
 Time  (median):     19.166 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   20.348 μs ±  27.253 μs  ┊ GC (mean ± σ):  4.52% ±  3.29%

     ▁▃▅▅▇███▇▇▅▄▃▂▁                                           ▂
  ▂▆██████████████████▇▇▆▆▅▅▄▆▆▇▆▇▇▇████▇▇▆▇▇▇▇▇▅▆▆▄▆▆▅▃▆▅▅▅▅▄ █
  18 μs         Histogram: log(frequency) by time      25.1 μs <

 Memory estimate: 61.17 KiB, allocs estimate: 484.
```

main
```
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  19.125 μs … 811.000 μs  ┊ GC (min … max): 0.00% … 94.50%
 Time  (median):     20.292 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   21.633 μs ±  29.285 μs  ┊ GC (mean ± σ):  4.93% ±  3.55%

         ▁▁▇█▆▂▁
  ▁▁▁▁▂▃▅███████▅▃▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  19.1 μs         Histogram: frequency by time         25.8 μs <

 Memory estimate: 60.22 KiB, allocs estimate: 485.
```

ararslan:
Will wait to trigger registration since I'd like to get #102 merged too

Issue #64, opened 2022-11-03
support multiple inheritance


jrevels:
when authoring a new schema version, it is not uncommon to want to inherit the required fields of two or more unrelated (in lineage) schema versions

This should be quite feasible/straightforward in Legolas-land, as long as we:

1. force the author to specify the order in which the multiple ancestors should be applied
2. support new syntax for this in our schema version identifiers, which captures the aforementioned ordering 
3. have some basic type checking to ensure that overlapping required fields aren't mutually incompatible, i.e. later-applied parents' required fields must either be entirely disjoint with earlier parents' required fields, or be valid subtypes of them

jrevels:
Note that it's currently pretty trivial to evaluate a predicate across a given schema version's ancestry:

```jl
function any_ancestor_matches_predicate(predicate, sv::Legolas.SchemaVersion)
    predicate(sv) && return true
    p = Legolas.parent(sv)
    isnothing(p) && return false
    predicate(p) && return true
    return any_ancestor_matches_predicate(predicate, p)
end
```

If we implement this, stuff like this will turn into a tree walk, which is just complicated enough that it might be worth providing some utilities around

jrevels:
> support new syntax for this in our schema version identifiers, which captures the aforementioned ordering


I think this should just be `[<lineage>, <lineage>, ...]`. Here's a (purposefully pretty complicated) example:

```
a > b > [c > d > [e,  f], g > [h, i > j], k > l > m] 
```

In Julia, the syntax would be:

```jl
@version FooV1 > [BarV1, BazV1] begin
    ....
end
```

PR #104, opened 2023-10-24
Allow Children of Children Versions of the same Schema

mattBrzezinski:
### Overview

Stemming from an internal discussion w/ @haberdashPI I was wanting to create a V2 schemaversion off of its V1 schemaversion. It wasn't clear _why_ this is disallowed? This PR loosens the constraint and allows for this situation.

```julia
using Legolas
using Legolas: @schema, @version

@schema "foo.bar" Bar

@version BarV1 begin
    a
end

@version BarV2 > BarV1 begin
    b
end
```

### DIFF
--- a/src/schemas.jl
+++ b/src/schemas.jl
@@ -836,8 +836,6 @@ macro version(record_type, declared_fields_block=nothing)
 
             if $Legolas.declared(schema_version) && $Legolas.declaration(schema_version) != schema_version_declaration
                 throw(SchemaVersionDeclarationError("invalid redeclaration of existing schema version; all `@version` redeclarations must exactly match previous declarations"))
-            elseif parent isa $Legolas.SchemaVersion && $Legolas.name(parent) == schema_name
-                throw(SchemaVersionDeclarationError("cannot extend from another version of the same schema"))
             elseif parent isa $Legolas.SchemaVersion && !($Legolas._has_valid_child_field_types($declared_field_names_types, $Legolas.declared_fields(parent)))
                 throw(SchemaVersionDeclarationError("declared field types violate parent's field types"))
             else
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -324,7 +324,6 @@ end
         @test_throws SchemaVersionDeclarationError("provided record type expression is malformed: BobV1 < DaveV1") @version(BobV1 < DaveV1, begin x end)
         @test_throws SchemaVersionDeclarationError("cannot have duplicate field names in `@version` declaration; received: $([:x, :y, :x, :z])") @version(ChildV2, begin x; y; x; z end)
         @test_throws SchemaVersionDeclarationError("cannot have field name which start with an underscore in `@version` declaration: $([:_X])") @version(ChildV2, begin x; X; _X end)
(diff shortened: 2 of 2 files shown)


omus:
This feels a little weird. I suspect this may be problematic with the addition of multiple inheritance.

Is the main motivation here to avoid having to copy the entirety of the previous schema version? If so I could see a convenience syntax to do that without having to utilize the inheritance system. For example with this PR if `BarV1` had a parent and in `BarV2` you wanted to modify the parent as well as copy the fields from `BarV1` you couldn't do that.

mattBrzezinski:
> This feels a little weird. I suspect this may be problematic with the addition of [multiple inheritance](https://github.com/beacon-biosignals/Legolas.jl/issues/64).
> 
> Is the main motivation here to avoid having to copy the entirety of the previous schema version? If so I could see a convenience syntax to do that without having to utilize the inheritance system. For example with this PR if `BarV1` had a parent and in `BarV2` you wanted to modify the parent as well as copy the fields from `BarV1` you couldn't do that.

Currently we have something like,

```
using Legolas: @schema, @version

@schema "foo.general" General

@version GeneralV1 begin
    general_field
end

@schema "foo.bar" BarV1

@version BarV1 > GeneralV1 begin
    bar_field
end
```

I want to introduce a new field to `BarV1`, however I do not want to tag a breaking change as we will be making a few of these changes in the future. To get around this the idea is to walk around semver for the package and introduce a new schema version,

```
@schema BarV2 > BarV1 begin
    new_field
end
```

From there we can tag a non-breaking release, warning users however to continue using `BarV1` until all work is complete on `BarV2`. Then deprecating `BarV1`, tagging that release. And then making a new release where `BarV1` is removed and usage is only through `BarV2`.

Making a breaking change isn't that problematic on its own, but current internal work will require modifying a few schema versions some of which requires some design work. I'm really trying to avoid the situation of making multiple breaking change releases in a short amount of time and causing some roughness with users of the package where `BarV1` is being defined.

ericphanson:
> And then making a new release where `BarV1` is removed and usage is only through `BarV2`.

This is not possible with this design:
```julia
julia> using Legolas

julia> @schema "foo.bar" Bar

julia> @version BarV1 begin
           a
       end

julia> @version BarV2 > BarV1 begin
           b
       end

julia> Legolas.identifier(BarV2SchemaVersion())
"foo.bar@2>foo.bar@1"
```
`Legolas.identifier` gets serialized into Legolas-produced tables, as the identifier for the schema, meaning to load such a table, the loader must have acess to the `foo.bar@2` and `foo.bar@1` schemas. Therefore, with this route, `foo.bar@1` can never be removed.

I second @omus in that you should just copy-paste, and we could automate that with a macro if really needed. That is, define
```julia
using Legolas
using Legolas: @schema, @version

@schema "foo.bar" Bar

@version BarV1 begin
    a
end

@version BarV2 begin
    a
    b
end
```
Then when you no longer need `BarV1`, you can remove it without touching the `BarV2` schema. (And there is no need for a breaking change if you don't need it; you can leave the `BarV1` schema for as long as you like).

> To get around this the idea is to walk around semver for the package and introduce a new schema version

BTW this is not getting around package semver, in fact, this is one reason why schemas have versions! I.e. this is the intended use of Legolas.jl.

PR #106, opened 2023-12-05
ensure deserialization is agnostic to serialization of field order

jrevels:
should be mergeable once https://github.com/apache/arrow-julia/pull/493 is released


Project.toml
```diff
@@ -1,18 +1,20 @@
 name = "Legolas"
 uuid = "741b9549-f6ed-4911-9fbf-4a1c0c97f0cd"
 authors = ["Beacon Biosignals, Inc."]
-version = "0.5.16"
+version = "0.6.0"
```
ararslan, line 4: Seems like this would constitute a bugfix rather than a breaking release?

jrevels: it introduces new higher compat bounds

ericphanson: That’s not considered breaking since Pkg will handle it and not install this version if you aren’t compatible

jrevels: okey dokey

jrevels:
```suggestion
version = "0.5.17"
```

PR #107, opened 2023-12-06
implement multiple inheritance

jrevels:
brought to you by: yesterday's MCO -> JFK plane ride

Still needs:

- proper deprecation of `Legolas.parent`
- updates to enable existing test suite to pass (+ fixes for any bugs caught by existing test suite)
- new tests to cover the multiple inheritance functionality and various edge cases that probably arise
- more comprehensive tour examples
- updates to existing documentation/docstrings
- a new section in the manual providing modeling guidance about when to use multiple inheritance vs. foreign key"-ish relationships (e.g. using junction records)
- an implementation of [a `has_matching_ancestor` utility function (since this is now more complicated for downstream users to hand-roll)](https://github.com/beacon-biosignals/Legolas.jl/issues/64#issuecomment-1313154264)

Would be very useful if folks could try this PR out in situ in downstream use cases and report back any encountered challenges / bugs / MWEs

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Legolas"
 uuid = "741b9549-f6ed-4911-9fbf-4a1c0c97f0cd"
 authors = ["Beacon Biosignals, Inc."]
-version = "0.5.16"
+version = "0.6.0"
 
 [deps]
 Arrow = "69666777-d1a9-59fb-9406-91d4454c9d45"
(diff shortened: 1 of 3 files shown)


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Legolas"
 uuid = "741b9549-f6ed-4911-9fbf-4a1c0c97f0cd"
 authors = ["Beacon Biosignals, Inc."]
-version = "0.5.16"
+version = "0.6.0"
```
jrevels, line 4: #106 will presumably land before this, in which case this should be bumped again

PR #108, opened 2023-12-18
Create function `record_merge`

omus:
Similar `Tables.rowmerge` but for `AbstractRecord`. We need to use a new function name as extending `Tables.rowmerge` would change the return type for existing `Table.rowmerge` callers. This function is purely for convenience and to avoid code duplication.


src/recordmerge.jl
```diff
@@ -0,0 +1,13 @@
+"""
+    recordmerge(record; fields_to_merge...)
```
ararslan, line 2: The symmetry with `rowmerge` is appealing but this is technically not a style guide compliant name. `record_merge` would work. Though would it make sense for this to instead be a method of `rowmerge` or even `merge`?

src/recordmerge.jl
```diff
@@ -0,0 +1,13 @@
...
+
+Return an `AbstractRecord` by merging the fields of `record` and `fields_to_merge` together.
```
ararslan, line 4:
```suggestion
Return a new `AbstractRecord` with the same schema as `record` by merging the fields of
`record` and `fields_to_merge` together using `Tables.rowmerge`.
```

src/recordmerge.jl
```diff
@@ -0,0 +1,13 @@
...
+
+See also: [`Tables.rowmerge`](@ref).
```
ararslan, line 6:
```suggestion
```
The Documenter `@ref` doesn't work for things that aren't included in the same output document.

src/recordmerge.jl
```diff
@@ -0,0 +1,13 @@
...
+"""
+function recordmerge(record::AbstractRecord; fields_to_merge...)
+    # Avoid using `typeof(record)` as can cause constructor failures with parameterized 
+    # record types.
+    R = Legolas.record_type(Legolas.schema_version_from_record(record))
```
ararslan, line 11:
```suggestion
    R = record_type(schema_version_from_record(record))
```
No need to qualify with the module name inside of the module since there are no locally conflicting names

src/recordmerge.jl
jrevels: > Though would it make sense for this to instead be a method of rowmerge or even merge?

i think this would make sense as a `rowmerge` method actually (should be docstring'd especially for the `AbstractRecord` method, of course)

src/recordmerge.jl
omus: Interesting. Surprised I haven't run into that before. I'd prefer to still have a mention here so maybe just without the link?

```suggestion

See also: `Tables.rowmerge`.
```

src/recordmerge.jl
omus: My bad. I forgot to remove this after upstreaming this change.

src/record_merge.jl
```diff
@@ -0,0 +1,12 @@
+"""
+    record_merge(record; fields_to_merge...)
+
+Return a new `AbstractRecord` with the same schema as `record` by merging the fields of
+`record` and `fields_to_merge` together using `Tables.rowmerge`.
+"""
+function record_merge(record::AbstractRecord; fields_to_merge...)
```
ararslan, line 7: Doesn't have to be part of this PR but it could be nice to have an additional method that works analogously to `rowmerge`, then this method could simply call that one:
```julia
function record_merge(record::AbstractRecord, other, others...)
    R = record_type(schema_version_from_record(record))
    return R(Tables.rowmerge(record, other, others...))
end

record_merge(record::AbstractRecord; fields...) = record_merge(record, (; fields...))
```

omus: I did notice that other method but decided to not introduce it as I personally didn't have a need for it. Also, we may want to restrict other rows to the same `AbstractRecord` subtype and avoid updating non-missing values with `missing`.

omus: The `missing` thing seems especially problematic as the keyword method allows you to set a field to `missing` but if we did the same with `record_merge(record::T, other::T)` you'll end up just with `other`.

omus:
Any more comments?

jrevels (approved):
LGTM - one last request before merging @omus , can you add `record_merge` to the list of API functions in the docs? so that the docstring shows up in the generated doc page

src/record_merge.jl
```diff
@@ -0,0 +1,12 @@
+"""
+    record_merge(record; fields_to_merge...)
+
+Return a new `AbstractRecord` with the same schema as `record` by merging the fields of
+`record` and `fields_to_merge` together using `Tables.rowmerge`.
+"""
```
jrevels, line 6:
```suggestion
"""
    record_merge(record::AbstractRecord; fields_to_merge...)

Return a new `AbstractRecord` with the same schema version as `record`, whose fields
are computed via `Tables.rowmerge(record; fields_to_merge...)`. The returned record is 
constructed by passing these merged fields to the `AbstractRecord` constructor matching 
the type of the input `record`.
"""
```

just some additional clarity for callers

ararslan (approved):

omus:
I noticed this missing reference when building the docs so I'll remove that while I'm at it:
```
┌ Warning: undefined binding 'Legolas.assign_to_table_metadata!' in `@docs` block in src/index.md:45-53
│ ```@docs
│ Legolas.lift
│ Legolas.construct
│ Legolas.record_merge
│ Legolas.assign_to_table_metadata!
│ Legolas.gather
│ Legolas.locations
│ Legolas.materialize
│ ```
└ @ Documenter.Expanders ~/.julia/packages/Documenter/PLD7m/src/Expanders.jl:298
```

omus:
... and took care of these too:
```
┌ Warning: no doc found for reference '[`@version`](@ref)' in src/upgrade.md.
└ @ Documenter.CrossReferences ~/.julia/packages/Documenter/PLD7m/src/CrossReferences.jl:160
┌ Warning: 1 docstring potentially missing:
│
│     Legolas.tobuffer :: Tuple
└ @ Documenter.DocChecks ~/.julia/packages/Documenter/PLD7m/src/DocChecks.jl:63
```

ararslan (approved):

PR #112, opened 2024-03-31
Move ConstructionBase dependency to an extension on 1.9+

ararslan:
The dependency on ConstructionBase serves to provide a few convenience methods for users of packages like Accessors; it doesn't provide any functionality used directly by Legolas. This makes it a perfect candidate for a package extension.

Package extensions are new in Julia 1.9, and Legolas currently supports Julia 1.6. That means that we can't fully remove ConstructionBase as a direct dependency yet, but we can restructure things to make that as easy as possible once we drop support for earlier Julia versions.

### DIFF
--- a/ext/LegolasConstructionBaseExt.jl
+++ b/ext/LegolasConstructionBaseExt.jl
@@ -1,0 +1,32 @@
+module LegolasConstructionBaseExt
+
+using ConstructionBase
+using Legolas
+
+using Legolas: AbstractRecord
+
+# We need a bit of extra work to integrate with ConstructionBase for pre-1.7 due
+# to the overload of `propertynames(::Tables.AbstractRow)`. We could overload
+# `check_properties_are_fields` but that's not part of the public API, so this
+# is safer.
+if VERSION < v"1.7"
+    ConstructionBase.getproperties(r::AbstractRecord) = NamedTuple(r)
+
+    # This is largely copy-paste from `ConstructionBase.setproperties_object`:
+    # https://github.com/JuliaObjects/ConstructionBase.jl/blob/cd24e541fd90ab54d2ee12ddd6ccd229be9a5f1e/src/ConstructionBase.jl#L211-L218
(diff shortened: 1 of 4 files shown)


ext/LegolasConstructionBaseExt.jl
```diff
@@ -0,0 +1,32 @@
...
+    function ConstructionBase.setproperties(r::R, patch::NamedTuple) where {R<:AbstractRecord}
+        nt = getproperties(r)
+        nt_new = merge(nt, patch)
+        ConstructionBase.check_patch_properties_exist(nt_new, nt, r, patch)
+        args = Tuple(nt_new)  # old Julia inference prefers if we wrap in `Tuple`
+        return constructorof(R)(args...)
+    end
+end
+
+function ConstructionBase.constructorof(::Type{<:R}) where {R<:AbstractRecord}
```
omus, line 26:
```suggestion
function ConstructionBase.constructorof(::Type{R}) where {R<:AbstractRecord}
```

src/Legolas.jl
```diff
@@ -10,4 +9,11 @@ include("schemas.jl")
 include("tables.jl")
 include("record_merge.jl")
 
+# TODO: Once we require Julia 1.9 or later at a minimum, we can remove this as well as
+# all entries in the Project.toml `[deps]` section that are also listed in `[weakdeps]`.
+if !isdefined(Base, :get_extension)
+    include(joinpath(dirname(@__DIR__), "ext", "LegolasConstructionBaseExt.jl"))
```
omus, line 15: A little obfuscated
```suggestion
    include(joinpath(@__DIR__(), "..", "ext", "LegolasConstructionBaseExt.jl"))
```

omus:
Needs a version bump

ext/LegolasConstructionBaseExt.jl
ararslan: Ah nice catch, copy-pasta artifact from https://github.com/beacon-biosignals/Legolas.jl/blob/main/src/schemas.jl#L631

Issue #20, opened 2021-08-10
implement basic table-level constraints + table/value-level validation


jrevels:
The most basic of which is a primary key relationship. Note that we actually originally decided not to have constraints in Legolas simply because we figured spelling e.g. `allunique(my_table.my_primary_key_column)` was sufficient (and likewise for `issubset` to test foreign key relations), but given that this is already [emerging as useful downstream functionality](https://github.com/beacon-biosignals/Onda.jl/pull/98/files#r686325162), I think we might as well just have fully generic versions supported by the package.

jrevels:
ref other packages that facilitate key handling

- https://github.com/JuliaData/IndexedTables.jl
- https://github.com/invenia/KeyedFrames.jl
- https://github.com/mcabbott/AxisKeys.jl

jrevels:
FWIW, my plan here API-wise is for `Legolas.validate` to basically validate "as much stuff as it can" given the input, then expect the caller to ensure that the input is the "actual thing they want to validate." This approach seems intuitive + appropriately caller-driven to me.

So:

- `Legolas.validate(table_schema::Tables.Schema, schema::Legolas.Schema)` only validates the schema (already implemented)
- `Legolas.validate(table, schema::Legolas.Schema)` will actually validate the contents of `table` as much as possible against `schema`
- `Legolas.validate(table)` will actually validate that `table` contains an extractable `Legolas.Schema`, and then call `Legolas.validate(table, schema::Legolas.Schema)` (already implemented)

jrevels:
wonder if there should be a separate notion of "check constraints" for schema authors to express besides just normal per-field RHS stuff. 

could imagine extending the macro:

```jl
@row(name, 
     fields..., 
     @check begin
         ...
     end)
``` 

then the check constraints could be invoked as part of one of the `validate` blocks (as well as construction)

jrevels:
I wonder how crazy it would be to add actual foreign key relations to Legolas:

```jl
@row(name, 
     fields..., 
     @foreign("other-schema@2", field), 
     @check begin
         ...
     end)
```

we'd be entering full relational IDL territory...things could get wacky. For example, would we allow composite foreign keys?

jrevels:
xref #69 for a solid sketch of what convenient record-level check constraints could look like

PR #119, opened 2024-06-11
Add constraint support

### LINKED ISSUE #69: `@check` convenience syntax for record/field-level value constraints
Split out from #20. With Legolas v0.5 (ref #54), we have a much nicer disciplined stance on validation-by-construction, so the path forward for convenient record-level check constraints is much simpler. 

We should define a simple helper `@check` that can be used in the following manner:

```jl
@version FooV1 begin
    x::String # unchecked
    @check(y::Int, y > 0)
    @check(z::Any, a_predicate(z) && another_predicate(z, x))
    @check(k::Int = clamp(k, 1, 10), 
           some_other_predicate(k))
    @check begin
        # record-level predicate(s)
    end
end
```

where the field-level `@check` expansion goes from statements like

```jl
@version FooV1 begin
    ⋮
    @check(f::T = $rhs, $predicate)
    ⋮
end
```

...to statements like:

```jl
f::T = $rhs
$predicate || throw(CheckConstraintError(:FooV1, :f, f, $stringified_predicate))
```

...where `CheckConstraintError` might look like:

```jl
struct CheckConstraintError <: Exception
    record_name::Symbol
    field_name::Symbol
    predicate::String
    received::Any
end

function Base.showerror(io::IO, e::CheckConstraintError)
    print(io, """
              Value received for $(e.field_name) of $(e.record_name): $(e.received)
              Violates `@check` constraint: $(e.predicate)
              """)
end
```

omus:
Adds basic constraint support for Legolas schemas so you can define constraints on the relationships between fields. Previously, this could be worked around by adding such logic into a field:

```julia

julia> @schema "old.constraint" OldConstraint

julia> @version OldConstraintV1 begin
           a
           b = begin
               @assert a == b
               b
           end
       end

julia> OldConstraintV1(; a=1, b=2)
ERROR: ArgumentError: Invalid value set for field `b` (2)
Stacktrace:
 [1] OldConstraintV1(; a::Int64, b::Int64)
   @ Main ~/.julia/dev/Legolas/src/schemas.jl:545
 [2] top-level scope
   @ REPL[4]:1

caused by: AssertionError: a == b
Stacktrace:
 [1] macro expansion
   @ ./REPL[3]:4 [inlined]
 [2] OldConstraintV1(; a::Int64, b::Int64)
   @ Main ~/.julia/dev/Legolas/src/schemas.jl:580
 [3] top-level scope
   @ REPL[4]:1
```

Unfortunately, the old approach would produce an extra `ArgumentError` which could create confusion. Additionally, when utilizing more complicated constructor logic for `b` these kinds of constraints can be less clear than declaring them separately like so:

```julia

julia> @schema "new.constraint" NewConstraint

julia> @version NewConstraintV1 begin
           a
           b
           @check a == b
       end

julia> NewConstraintV1(; a=1, b=2)
ERROR: Legolas.CheckConstraintError: a == b
Stacktrace:
 [1] macro expansion
   @ ./REPL[3]:4 [inlined]
 [2] NewConstraintV1(; a::Int64, b::Int64)
   @ Main ~/.julia/dev/Legolas/src/schemas.jl:619
 [3] top-level scope
   @ REPL[4]:1
```

Closes #69

### DIFF
--- a/src/schemas.jl
+++ b/src/schemas.jl
@@ -787,12 +787,10 @@ macro version(record_type, declared_fields_block=nothing)
             if f isa LineNumberNode
                 continue
             elseif f isa Expr && f.head === :macrocall && f.args[1] === Symbol("@check")
-                # Avoids having to import `@check`
-                f.args[1] = Expr(:., :Legolas, QuoteNode(f.args[1]))
                 # Expand `@check` macro here so that we can reliably see the location of
                 # the user define `@check` when it fails. Ideally `Meta.replace_sourceloc!`
                 # would do this for us.
-                constraint_expr = Base.macroexpand(__module__, f)
+                constraint_expr = Base.macroexpand(Legolas, f)
                 if f.args[2] isa LineNumberNode
                     constraint_expr = Expr(:block, f.args[2], constraint_expr)
                 end
--- a/src/schemas.jl
+++ b/src/schemas.jl
@@ -787,6 +787,8 @@ macro version(record_type, declared_fields_block=nothing)
             if f isa LineNumberNode
                 continue
             elseif f isa Expr && f.head === :macrocall && f.args[1] === Symbol("@check")
+                # Avoids having to import `@check`
+                f.args[1] = Expr(:., :Legolas, QuoteNode(f.args[1]))
                 # Expand `@check` macro here so that we can reliably see the location of
                 # the user define `@check` when it fails. Ideally `Meta.replace_sourceloc!`
                 # would do this for us.
--- a/src/schemas.jl
+++ b/src/schemas.jl
@@ -786,9 +786,9 @@ macro version(record_type, declared_fields_block=nothing)
         for f in declared_fields_block.args
             if f isa LineNumberNode
                 continue
-            elseif f isa Expr && f.head === :macrocall && f.args[1] === Symbol("@assert")
-                # Expand `@assert` macro here so that we can reliably see the location of
-                # the user define `@assert` when it fails. Ideally `Meta.replace_sourceloc!`
+            elseif f isa Expr && f.head === :macrocall && f.args[1] === Symbol("@check")
+                # Expand `@check` macro here so that we can reliably see the location of
+                # the user define `@check` when it fails. Ideally `Meta.replace_sourceloc!`
                 # would do this for us.
                 constraint_expr = Base.macroexpand(__module__, f)
                 if f.args[2] isa LineNumberNode
--- a/src/schemas.jl
+++ b/src/schemas.jl
@@ -787,10 +787,9 @@ macro version(record_type, declared_fields_block=nothing)
             if f isa LineNumberNode
                 continue
             elseif f isa Expr && f.head === :macrocall && f.args[1] === Symbol("@check")
-                # Expand `@check` macro here so that we can reliably see the location of
-                # the user define `@check` when it fails. Ideally `Meta.replace_sourceloc!`
-                # would do this for us.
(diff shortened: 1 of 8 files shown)


omus:
GitHub actions is currently having an outage

jrevels:
xref https://github.com/beacon-biosignals/Legolas.jl/issues/69

I'd like to use `@check` for this instead of `@assert`; want to avoid teaching users that it's literally the same as Julia's built-in`@assert`, as we might add more sugar/behaviors later to this (e.g. https://github.com/beacon-biosignals/Legolas.jl/issues/20)

omus:
> I'd like to use `@check` for this instead of `@assert`; want to avoid teaching users that it's literally the same as Julia's built-in`@assert`, as we might add more sugar/behaviors later to this (e.g. #20)

Good call. I started off with `@assert` just to prove this was a viable approach. I will update this PR to use `@check`

test/runtests.jl
```diff
@@ -817,3 +818,72 @@ end
...
+    catch e
+        @test e isa CheckConstraintError
+        @test e.predicate == :(a == b)
+    end
+
+    # For exceptions that occur during processing constraints its convenient to include the
+    # location of the `@check` in the stacktrace.
+    try
+        ConstraintV1(; a=1, b=missing)  # Fails when running check `a == b`
+        @test false
```
kleinschmidt, line 876: ??

src/schemas.jl
```diff
@@ -778,8 +781,26 @@ macro version(record_type, declared_fields_block=nothing)
...
+            if f isa LineNumberNode
+                continue
+            elseif f isa Expr && f.head === :macrocall && f.args[1] === Symbol("@check")
+                # Expand `@check` macro here so that we can reliably see the location of
+                # the user define `@check` when it fails. Ideally `Meta.replace_sourceloc!`
+                # would do this for us.
+                constraint_expr = Base.macroexpand(Legolas, f)
+                if f.args[2] isa LineNumberNode
+                    constraint_expr = Expr(:block, f.args[2], constraint_expr)
+                end
```
kleinschmidt, line 796: this is pretty opaque to me, add a comment explaining why this is necessary?

kleinschmidt: we don't want to just strip out the line number nodes right, since that's what will give the nice error w/ teh correct definition location when the check fails, right?

kleinschmidt: ah wait I think I see: we're creating a new `begin...end` with the line number node in the macrocall manually spliced in (if there is one).

kleinschmidt (commented):
I'm generally good with this.  I think a bit of explaining about the line number node manipulations would be helpful for future maintainers.

How reliable is the line-number manipulation across julia versions?  I have a vague sense that it might uh not be the most reliable thing but I have nothing to back that up...

test/runtests.jl
omus: Ensures that if `ConstraintV1(; a=1, b=missing)` does not thrown an exception that a test failure is reported.

kleinschmidt: ah okay, could use a comment too.

src/schemas.jl
omus: I've updated the comment to something which I think is clearer

kleinschmidt (approved):

omus:
I'll be waiting until https://github.com/beacon-biosignals/Legolas.jl/pull/123 is merged before merging this.

omus:
#123 has been merged. I need to update the version here and finish some downstream testing I started.

PR #122, opened 2024-08-28
Store package names in arrow metadata

### LINKED ISSUE #46: Feature request: include package name or package url in arrow metadata
Since Legolas schemas are defined in code, if you just have a table, `Legols.read` will tell you the schema isn't defined and you need to load the package, but it won't tell you *which* package (brought up by @a-cakir in the engineering roundtable). Namespacing can help, but it could be more clear if we actually just put the julia module name where the `@row` was defined, and/or the package name, and/or the repo url. We could put this in the Arrow metadata and use it to give a more precise error message.

ericphanson:
for more informative unknown schema errors

closes #46 

Here I chose to only store this information when the schema was defined in a package (rather than in Main or in a script or something), since it's only in that case we can reliably make a suggestion to load that package.


src/tables.jl
```diff
@@ -165,6 +171,14 @@ function read(io_or_path; validate::Bool=true)
                                              via `Legolas.read`; is it missing the expected custom metadata and/or the
                                              expected \"$LEGOLAS_SCHEMA_QUALIFIED_METADATA_KEY\" field?
                                              """))
+
+        provider = extract_metadata(table, LEGOLAS_SCHEMA_PROVIDER_METADATA_KEY)
+        # If we don't have the schema defined in our session (i.e. `Legolas.schema_provider` is `nothing`), 
+        # but we do have a hint of where the schema was defined via the metadata, then throw an informative
+        # error. If we don't error now, we will throw an `UnknownSchemaVersionError` with less information later.
+        if Legolas.schema_provider(Val(Legolas.name(sv))) === nothing && provider !== nothing
```
kleinschmidt, line 179: why can't we just dispatch on the schema version itself?  seems weird to introduce `Val`-based dispatch here when we have the sv already from the parsed identifier in the metadata..

src/schemas.jl
```diff
@@ -165,6 +185,24 @@ written via [`Legolas.write`](@ref).
 """
 identifier(sv::SchemaVersion) = throw(UnknownSchemaVersionError(sv))
 
+"""
+    Legolas.schema_provider(::Val{schema_name}) where schema_name
+
+Returns a `Symbol` corresponding to the package which defines the schema, if known.
+Otherwise returns `nothing`.
+"""
+schema_provider(::Val) = nothing
```
kleinschmidt, line 194:
```suggestion
schema_provider(::SchemaVersion) = nothing
```

src/schemas.jl
```diff
@@ -375,12 +413,13 @@ macro schema(schema_name, schema_prefix)
...
     return quote
         # This approach provides some safety against accidentally replacing another module's schema's name,
-        # without making it annoying to reload code/modules in an interactice development context.
+        # without making it annoying to reload code/modules in an interactive development context.
         m = $Legolas._schema_declared_in_module(Val(Symbol($schema_name)))
         if m isa Module && string(m) != string(@__MODULE__)
             throw(ArgumentError(string("A schema with this name was already declared by a different module: ", m)))
         else
             $Legolas._schema_declared_in_module(::Val{Symbol($schema_name)}) = @__MODULE__
+            $Legolas.schema_provider(::Val{Symbol($schema_name)}) = $Legolas.defining_package(@__MODULE__)
```
kleinschmidt, line 422: oh i see, this a parallel to `_schema_declared_in_module`.  doesn't that do what you want here?

kleinschmidt: I actually think it might not: even if the _schema_ is defined, the _version_ might not, which is just as bad as far as dealing with the arrow metadata is concerned...

src/tables.jl
ericphanson: I did it in the `@schema` macro rather than the `@version` macro so it's independent of version, but I could move it to `@version` and use the full sv

src/schemas.jl
ericphanson: yeah, I wanted the defining package as a symbol, which is slightly different than `_schema_declared_in_module` but could be derived from it. But I agree, probably I should move this to the `@version` macro and use the full SV

Project.toml
```diff
@@ -31,11 +32,12 @@ Accessors = "7d9f7c33-5ae7-4f3b-8dc6-eff91059b697"
...
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [targets]
-test = ["Accessors", "Aqua", "Compat", "DataFrames", "Test", "UUIDs"]
+test = ["Accessors", "Aqua", "Compat", "DataFrames", "Pkg", "Test", "UUIDs"]
 
 [weakdeps]
-ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
+ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
```
ericphanson, line 43:
```suggestion
ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"

```

ericphanson:
good call @kleinschmidt, I've switched this to use the schema-version instead of the schema name (defined in the `@version` call rather than the `@schema` call)

omus:
@ericphanson you probably also want to include the version of the package as well since older or newer versions may be incompatible.

Project.toml
```diff
@@ -18,9 +18,10 @@ ArrowTypes = "2.3"
 Compat = "3.34, 4"
 ConstructionBase = "1.5.7"
 DataFrames = "1"
+Pkg = "<0.0.1, 1"
```
omus, line 21: Is `0.0.0` not a thing?

src/schemas.jl
```diff
@@ -98,8 +98,12 @@ end
 
 struct UnknownSchemaVersionError <: Exception
     schema_version::SchemaVersion
+    schema_provider_name::Union{Missing, Symbol}
+    schema_provider_version::Union{Missing, VersionNumber}
```
omus, line 102: Looks like we don't have lint checking on this repo. I think the these don't have spaces
```suggestion
    schema_provider_name::Union{Missing,Symbol}
    schema_provider_version::Union{Missing,VersionNumber}
```

src/schemas.jl
```diff
@@ -110,13 +114,33 @@ function Base.showerror(io::IO, e::UnknownSchemaVersionError)
               This generally indicates that this schema has not been declared (i.e.
               the corresponding `@schema` and/or `@version` statements have not been
               executed) in the current Julia session.
+              """)
+    println(io)
+
+    if !ismissing(e.schema_provider_name)
+        provider_string = string(e.schema_provider_name)
+        if !ismissing(e.schema_provider_version)
+            provider_string = string(provider_string, " (version: ", e.schema_provider_version, ")")
```
omus, line 123:
```suggestion
            provider_string *= string(" (version: ", e.schema_provider_version, ")")
```

src/schemas.jl
```diff
@@ -165,6 +189,25 @@ written via [`Legolas.write`](@ref).
 """
 identifier(sv::SchemaVersion) = throw(UnknownSchemaVersionError(sv))
 
+"""
+    Legolas.schema_provider(::SchemaVersion)
+
+Returns a NamedTuple with keys `name` and `version`. The name is a `Symbol` corresponding to the package which defines the schema version, if known; otherwise `nothing`. Likewise the `version` is a `VersionNumber` or `nothing`.
+"""
+schema_provider(::SchemaVersion) = (; name=nothing, version=nothing)
```
omus, line 197: Why do we switch to `nothing` here?

src/tables.jl
```diff
@@ -213,11 +228,26 @@ function write(io_or_path, table, sv::SchemaVersion; validate::Bool=true,
...
+    provider_name_metadata = LEGOLAS_SCHEMA_PROVIDER_NAME_METADATA_KEY => string(provider_name)
+    provider_version_metadata = LEGOLAS_SCHEMA_PROVIDER_VERSION_METADATA_KEY => string(provider_version)
     if isnothing(metadata)
         metadata = (schema_metadata,)
+        if !isnothing(provider_name)
+            metadata = (metadata..., provider_name_metadata)
+            if !isnothing(provider_version)
+                metadata = (metadata..., provider_version_metadata)
+            end
+        end
```
omus, line 241: Should we add this to `schema_metadata` instead?

test/runtests.jl
```diff
@@ -3,6 +3,31 @@ using Legolas, Test, DataFrames, Arrow, UUIDs
...
+# This test set goes before we load `TestProviderPkg`
+@testset "#46: Informative errors when reading unknown schemas from packages" begin
+    err = Legolas.UnknownSchemaVersionError(Legolas.SchemaVersion("test-provider-pkg.foo", 1), :TestProviderPkg, v"0.1.0")
+    @test_throws err Legolas.read("test_provider_pkg.arrow")
+    @test contains(sprint(Base.showerror, err), "TestProviderPkg")
+
+    # Let's test some more error printing while we're here; if we did not have the VersionNumber
+    # (e.g. since the table was generated on Julia pre-1.9), we should still print a reasonable message:
+    err = Legolas.UnknownSchemaVersionError(Legolas.SchemaVersion("test-provider-pkg.foo", 1), :TestProviderPkg, missing)
+    @test contains(sprint(Base.showerror, err), "TestProviderPkg")
```
omus, line 17: We should use `Arrow.Table` to load the table without the package to ensure that the advice we give is accurate

Project.toml
ericphanson: No, Pkg doesn’t understand it IIRC

src/schemas.jl
ericphanson: I switched to missing elsewhere since that’s what `lift` returns and I wanted to use `lift`

src/tables.jl
ericphanson: I’m not sure tuples are supported if that’s what you mean

test/runtests.jl
```diff
@@ -3,6 +3,31 @@ using Legolas, Test, DataFrames, Arrow, UUIDs
...
+# Now load the package, and verify we can write the tables with this metadata
+Pkg.develop(; path=joinpath(@__DIR__, "TestProviderPkg"))
+using TestProviderPkg
+
+@testset "#46: Writing informative metadata about packages providing schemas" begin
+    table = [TestProviderPkg.FooV1(; a=1)]
+    Legolas.write("test_provider_pkg.arrow", table, TestProviderPkg.FooV1SchemaVersion())
+    table = Legolas.read("test_provider_pkg.arrow")
+    v = Legolas.extract_metadata(table, Legolas.LEGOLAS_SCHEMA_PROVIDER_NAME_METADATA_KEY)
+    @test v == "TestProviderPkg"
```
omus, line 29: Add a test for the version key?

src/schemas.jl
```diff
@@ -165,6 +189,25 @@ written via [`Legolas.write`](@ref).
...
+# shadow `pkgversion` so we don't fail on pre-1.9
+pkgversion(m::Module) = isdefined(Base, :pkgversion) ? Base.pkgversion(m) : nothing
+
+# Used in the implementation of `schema_provider`.
+function defining_package_version(m::Module)
+    rootmodule = Base.moduleroot(m)
+    # Check if this module was defined in a package.
+    # If not, return `nothing`
+    path = pathof(rootmodule)
+    path === nothing && return (; name=nothing, version=nothing)
```
omus, line 207: Comment above makes it sound like this should be:
```suggestion
    path === nothing && return nothing
```

omus (approved):
Some minor code suggestions. Please bump the version to `0.5.22`

test/runtests.jl
kleinschmidt: agreed

src/schemas.jl
kleinschmidt: needs to be consistent with the return value described in the docstring of `schema_provider` though, which is as it is here...

src/tables.jl
```diff
@@ -213,11 +228,26 @@ function write(io_or_path, table, sv::SchemaVersion; validate::Bool=true,
...
     else
         metadata = Set(metadata)
         push!(metadata, schema_metadata)
+        if !isnothing(provider_name)
+            push!(metadata, provider_name_metadata)
+            if !isnothing(provider_version)
+                push!(metadata, provider_version_metadata)
+            end
+        end
     end
```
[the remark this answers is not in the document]
kleinschmidt: I _think_ I agree with this, and it's easier for me to follow at least

kleinschmidt (commented):
mostly agree with @omus review and have left comments where I don't.  needs a version bump.

src/schemas.jl
ericphanson: fixed the comment

src/tables.jl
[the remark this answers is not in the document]
ericphanson: switched to a version of this

ericphanson:
sorry for the delay, I think I've addressed all the review feedback. I'll plan to merge in a day or two if there's no more review changes.

src/schemas.jl
```diff
@@ -110,13 +114,33 @@ function Base.showerror(io::IO, e::UnknownSchemaVersionError)
               This generally indicates that this schema has not been declared (i.e.
               the corresponding `@schema` and/or `@version` statements have not been
               executed) in the current Julia session.
+              """)
+    println(io)
+
+    if !ismissing(e.schema_provider_name)
+        provider_string = string(e.schema_provider_name)
+        if !ismissing(e.schema_provider_version)
+            provider_string *= string(" (version: ", e.schema_provider_version, ")")
```
omus, line 123: Having `Example (version: v1.0.0)` seems a bit verbose. Using `Example v1.0.0` seems fine to me:
```suggestion
            provider_string *= string(" ", e.schema_provider_version)
```

src/schemas.jl
```diff
@@ -110,13 +114,33 @@ function Base.showerror(io::IO, e::UnknownSchemaVersionError)
...
+        end
+        print(io, """
+                The table's metadata indicates that the table was created with a schema defined in:
 
-              In practice, this can arise if you try to read a Legolas table with a
-              prescribed schema, but haven't actually loaded the schema definition
-              (or commonly, haven't loaded the dependency that contains the schema
-              definition - check the versions of loaded packages/modules to confirm
-              your environment is as expected).
+                $(provider_string)
```
omus, line 128: Indented this by 2-spaces:
```suggestion
                  $(provider_string)
```

omus (approved):

PR #124, opened 2024-09-05
Add check for all missing arguments to schema.

haberdashPI:
A common mistake in using Legolas is to call `MySchemaV1()` instead of `MySchemaV1SchemaVersion()`. Since calling `MySchemaV1()` is rarely intentional, this PR flags a warning when that method is called. The warning can be silenced by calling `MySchemaV1(Legolas.AllMissing())` to signal that, yes, you really do want a record with all missing fields.

### DIFF
--- a/src/schemas.jl
+++ b/src/schemas.jl
@@ -515,24 +515,21 @@ end
 
 _schema_version_from_record_type(::Nothing) = nothing
 
-abstract type MissingSignal end
-struct NotAllMissing <: MissingSignal end
-struct AllMissing <: MissingSignal end
+struct DefaultArgument end
+is_default(::Any) = false
+is_default(::DefaultArgument) = true
+default_to(::DefaultArgument, default) = default
+default_to(x, ::Any) = x
 
-function _check_all_missing(::AllMissing, schema_name; kwargs...)
-    if !all(ismissing, values(kwargs))
-        throw(ArgumentError("Expected all arguments passed to `$schema_name` to be missing."))
-    end
-end
-
-function _check_all_missing(::NotAllMissing, schema_name; kwargs...)
-    if all(ismissing, values(kwargs))
-        @warn "The arguments passed to `$schema_name` are all missing. "*
+function _check_empty_args(schema_name, args)
+    if all(is_default, args)
+        @warn "The were no arguments passed to `$schema_name`. "*
               "Are you sure you don't mean `$(schema_name)SchemaVersion`?"*
               " If this is what you really meant you can call "*
-              "`$schema_name(Legolas.AllMissing())` to avoid this warning."*
+              "`$schema_name(; arg1=missing, arg2=...)` to avoid this warning."*
               " Future (breaking) versions of Legolas may make this warning an error."
     end
+    return default_to.(args, missing)
 end
 
 # Note also that this function's implementation is allowed to "observe" `Legolas.declared_fields(parent)`
@@ -612,7 +609,7 @@ function _generate_record_type_definitions(schema_version::SchemaVersion, record
     end
 
     # generate `parent_record_application`
-    field_kwargs = [Expr(:kw, n, :missing) for n in keys(record_fields)]
+    field_kwargs = [Expr(:kw, n, :($Legolas.DefaultArgument())) for n in keys(record_fields)]
     parent_record_application = nothing
     parent = Legolas.parent(schema_version)
     if !isnothing(parent)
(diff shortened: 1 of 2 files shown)


ericphanson:
IMO there shouldn't be any warning if I do `MySchema(row_with_all_missing)` or `MySchema(; col1=missing, col2=missing)`. I could see the rationale for a literal invocation of `MySchema()` to warn though.

iamed2:
As an addendum to @ericphanson's statement, I think `MySchemaV1((;))` should not warn.

But I wonder if this is best solved in a method signature. Are there cases where it makes sense to have an identical method signature that differs only between `::SchemaVersion` and `::AbstractRecord` in one parameter, where the result would be different?

haberdashPI:
Not sure I follow @iamed2:

> But I wonder if this is best solved in a method signature. 

Are you saying that downstream consumers should be able to handle both `MySchemaV1` and `MySchemaV1SchemaVersion`? 

I think the issue is that you can get an error when using `MySchemaV1()` (because one of the fields is not allowed to be missing). This is often confusing, and part of the role of making the method warn/error would be to make the message users get more legible.

ericphanson:
I think @iamed2 is suggesting method authors should take care to avoid this problem by type annotations in their methods signatures. I agree that would be good too, but I think suitably restricted this could be a reasonable fallback. FWIW a culprit of bad method signatures in this department is a particular closed source database client.

haberdashPI:
Okay, thanks for clarifying. Sounds like I need to adjust this PR by:

- remove the check for all missing keyword arguments
- replace with a check for an empty list of keyword arguments

haberdashPI:
As a rembinder to myself: still todo is to make a method for `MySchemaV1((;))` so that it doesn't error (probably by just passing `missing` instead of `Legoals.DefaultArgument`).

ericphanson:
Hm, is it feasible to add a second constructor with no arguments and make that one warn? That might be simpler, but I suppose it’s only valid if every field is nullable

haberdashPI:
> Hm, is it feasible to add a second constructor with no arguments and make that one warn? That might be simpler, but I suppose it’s only valid if every field is nullable

Yes, that sounds simpler / better. Right now I think it would be equivalent behavior to do that way. If we want the error about a given field not allowing missing to show up *without* a warning we'd have to use the inner constructor, but I actually think the warning remains helpful in that case, since the original mistake that motivates this change is sort of agnostic to the constraints on the fields.

haberdashPI:
> Hm, is it feasible to add a second constructor with no arguments and make that one warn? That might be simpler, but I suppose it’s only valid if every field is nullable

It's funny because I wrote a reply saying that yes, we could simplify this, and then I remembered why we can't.

The method `function $R(; $(field_kwargs...))` overlaps with `R()`, so by defining an empty constructor, we're actually defining the same method twice. Probably good to write a comment about that rationale.

kleinschmidt (changes requested):
I think this is probably not worth the added complexity to what is pretty complex code that is hard to read/reason about enough already.  If downstream consumers require an `AbstractRecord` then they should say so, or, alternatively, if they cannot handle a `SchemaVersion` then they can define a method that throws and error when it gets one (to allow generic record handling with an `::Any` method).  I dunno if that was what @iamed2 was suggesting but if it is then I agree with them 😁

haberdashPI:
> If downstream consumers require an AbstractRecord then they should say so, or, alternatively, if they cannot handle a SchemaVersion then they can define a method that throws and error when it gets one (to allow generic record handling with an ::Any method).

I think that makes sense from the perspective of someone developing a package.

When the downstream consumer is a migration or script, this just feels like an annoying foot-gun to run into repeatedly. It *is* arguably a *dumb* mistake (😆 ), but I think there's a lot of value in legible error messages.

I guess your point is that there might be more foot-guns hidden in the effort to write this error checking and keep it working. But this doesn't change any of the extant behavior, it warning. So I don't think I follow the maintenance argument.

src/schemas.jl
```diff
@@ -638,6 +674,7 @@ function _generate_record_type_definitions(schema_version::SchemaVersion, record
         end
         outer_constructor_definitions = quote
             $outer_constructor_definitions
+            $R{$(type_param_names...)}(::NamedTuple) where {$(type_param_names...)} = $R{$(type_param_names...)}($Legolas.DontCheckArgs())
             $R{$(type_param_names...)}(row) where {$(type_param_names...)} = $R{$(type_param_names...)}(; $(kwargs_from_row...))
```
kleinschmidt, line 678: I don't undersatnd what's happening here.  Doesn't this change the existing behavior (the case when `row` is a `NamedTuple`)?  if it _doesn't_ change the behavior, it at the very least needs a comment explaining why this is reasonable.

src/schemas.jl
```diff
@@ -607,13 +632,20 @@ function _generate_record_type_definitions(schema_version::SchemaVersion, record
 
     # generate `inner_constructor_definitions` and `outer_constructor_definitions`
     R = record_type_symbol
-    kwargs_from_row = [Expr(:kw, n, :(get(row, $(Base.Meta.quot(n)), missing))) for n in keys(record_fields)]
+    kwargs_from_row = [Expr(:kw, n, :(get(row, $(Base.Meta.quot(n)), $Legolas.DefaultArgument()))) for n in keys(record_fields)]
```
kleinschmidt, line 635: does this affect existing stack traces when a malformed input gets included?

kleinschmidt (commented):
> So I don't think I follow the maintenance argument

the maintenance issue is that even knowing what this is supposed to, being pretty familiar with the extremely cursed innards of legolas, and seeing exactly the diff of lines that were changed, and staring at it hard for 15 minutes, I'm still struggling to undersatnd exactly what's going on here.

I know from bitter experience that every extra layer of indirection in this kinda code means it's more likely that bugs will sneak in with innocuous looking changes, because there's so many hidden connections between things.  It's definitely a hunch, but my strong hunch is that this will cause more problems in the future than it fixes, and the "oops I used an empty record in place of a schema version" failure is a pretty innocuous one IMO.  Annoying?  Yes.  But like @iamed2 said, I think it's pretty reasonable even in interactive/script situations to use say `Dict{SchemaVersion,Any}` instead of `Dict` when you want to use a schema version as a key.

kleinschmidt:
I'll also add that strictly from a usability perspective, it solves one relatively rare and benign issue (using record instance where schema version instance was intended) by introducing another usability issue (inconsistent handling of ommitted vs. explicit `missing` kwargs in teh "completely empty" case).

haberdashPI:
Thanks for the explanation and feedback @kleinschmidt, the extra detail is helpful. If I somehow think of a simpler way to do this, I'll open a new PR.

PR #127, opened 2024-10-31
fix @check docstring

palday:
Based on actual use cases and examples in the tour, "follow" was meant, not "proceed"


kleinschmidt (approved):
postcede?

PR #129, opened 2024-11-19
Adjust to upcoming Julia 1.12 change in implicit world age increment

Keno:
In 1.12, we'll be cleaning up and making consistent when world age increment happens. This package is currently relying on implicit world age increments that may go away in 1.12. See https://github.com/JuliaLang/julia/pull/56509. This PR adds appropriate annotations that the world age increment is required because the `_generate_*` invocations in the macro read the result of the previous eval. This PR strictly retains existing behavior. A potentially better solution may be to `invokelatest` the `_generate_*` functions, which would also allow this code to work at non-toplevel (which may nor may not be desirable).


kleinschmidt (approved):

Keno:
Could I get a version tagged that includes this PR? That way PkgEval will be happy and keep testing this package and its downstreams once the upstream PR merges.

ericphanson:
v0.5.23 is released with this change, ty!

PR #132, opened 2025-04-11
Add accepted_field_type for VersionNumber

kendal-s:
VersionNumber is a base type. Defined as a struct, it has a pretty gnarly typing. If we want other languages that can easily construct arrow tables that are compatible with Legolas then we want to allow strings to be accepted as VersionNumbers. This PR implements this change.

### DIFF
--- a/src/schemas.jl
+++ b/src/schemas.jl
@@ -305,7 +305,7 @@ accepted_field_type(sv::SchemaVersion, ::Type{<:Vector{T}}) where T = AbstractVe
 accepted_field_type(::SchemaVersion, ::Type{Vector}) = AbstractVector
 accepted_field_type(sv::SchemaVersion, ::Type{Union{T,Missing}}) where {T} = Union{accepted_field_type(sv, T),Missing}
 accepted_field_type(::SchemaVersion, ::Type{Missing}) = Missing
-accpeted_field_type(::SchemaVersion, ::Type{VersionNumber}) = Union{VersionNumber,AbstractString}
+accepted_field_type(::SchemaVersion, ::Type{VersionNumber}) = Union{VersionNumber,AbstractString}
 
 """
     Legolas.find_violation(ts::Tables.Schema, sv::Legolas.SchemaVersion)
--- a/src/schemas.jl
+++ b/src/schemas.jl
@@ -305,7 +305,7 @@ accepted_field_type(sv::SchemaVersion, ::Type{<:Vector{T}}) where T = AbstractVe
 accepted_field_type(::SchemaVersion, ::Type{Vector}) = AbstractVector
 accepted_field_type(sv::SchemaVersion, ::Type{Union{T,Missing}}) where {T} = Union{accepted_field_type(sv, T),Missing}
 accepted_field_type(::SchemaVersion, ::Type{Missing}) = Missing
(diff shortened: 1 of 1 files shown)


src/schemas.jl
```diff
@@ -305,6 +305,7 @@ accepted_field_type(sv::SchemaVersion, ::Type{<:Vector{T}}) where T = AbstractVe
 accepted_field_type(::SchemaVersion, ::Type{Vector}) = AbstractVector
 accepted_field_type(sv::SchemaVersion, ::Type{Union{T,Missing}}) where {T} = Union{accepted_field_type(sv, T),Missing}
 accepted_field_type(::SchemaVersion, ::Type{Missing}) = Missing
+accpeted_field_type(::SchemaVersion, ::Type{VersionNumber}) = Union{VersionNumber,String}
```
kleinschmidt:
```suggestion
accpeted_field_type(::SchemaVersion, ::Type{VersionNumber}) = Union{VersionNumber,AbstractString}
```

kleinschmidt (commented):
I wanna add a bit of additional context here; @kendal-s and I came up with this approach while trying to figure out how to inter-operate w/ a python-produced arrow table where we were storing python-generated `semver.Version` structs as `(; major, minor, patch)` named tuple and were trying to expand this to handle prerelease and build versions as well.  Because python `semver.Version` [stores the prerelease and build as strings](https://github.com/python-semver/python-semver/blob/d8813b67a2ec5d579365f51a6d8f96d02f3fcbc3/src/semver/version.py#L136-L140) but Julia's `VersionNumber` represents them as _tuples_ of strings/ints, it's pretty tricky to define a pyarrow schema that matches the layout that julia expects.

Julia `VersionNumber` constructor _does_ accept a string though, so saving the version as a stringified form is pretty painless.  Unfortunately, it doens't _just work_ in the sense of allowing a `String` to be passed to the record constructor for a field that expects a `VersionNumber`, but that's no different from the situation with `Symbol` which also has a field type override to also accept an `AbstractString`...

src/schemas.jl
```diff
@@ -305,6 +305,7 @@ accepted_field_type(sv::SchemaVersion, ::Type{<:Vector{T}}) where T = AbstractVe
 accepted_field_type(::SchemaVersion, ::Type{Vector}) = AbstractVector
 accepted_field_type(sv::SchemaVersion, ::Type{Union{T,Missing}}) where {T} = Union{accepted_field_type(sv, T),Missing}
 accepted_field_type(::SchemaVersion, ::Type{Missing}) = Missing
+accpeted_field_type(::SchemaVersion, ::Type{VersionNumber}) = Union{VersionNumber,AbstractString}
```
kendal-s:
```suggestion
accepted_field_type(::SchemaVersion, ::Type{VersionNumber}) = Union{VersionNumber,AbstractString}
```

omus:
I'd like to see some tests for this, specifically:

1. A test which performs a round-trip on a `VersionNumber`.
2. A test which shows Legolas is compatible with the Python Arrow format for version numbers.
3. A test which shows Legolas reading a `VersionNumber` from the old struct form.

kleinschmidt:
> A test which shows Legolas is compatible with the Python Arrow format for version numbers.

I don't think that's the intent here though.  We're using a string _because_ the internal representation is sufficiently different to make specifying the appropriate pyarrow struct pretty difficult.

PR #142, opened 2026-08-12
fix for tests on Julia 1.13

palday:
The tests relied on non-public API, which changed on 1.13. This updates the test to avoid that and updates CI a bit to improve coverage around Julia versions. Versioned Manifests are also now ignored.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Legolas"
 uuid = "741b9549-f6ed-4911-9fbf-4a1c0c97f0cd"
 authors = ["Beacon Biosignals, Inc."]
-version = "0.5.24"
+version = "0.5.25"
 
 [deps]
 Arrow = "69666777-d1a9-59fb-9406-91d4454c9d45"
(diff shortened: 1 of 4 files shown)


test/runtests.jl
```diff
@@ -919,7 +919,7 @@ end
     catch e
         @test e isa TypeError
 
-        bt = Base.process_backtrace(catch_backtrace())
+        bt = Base.process_backtrace(stacktrace(catch_backtrace()))
```
kleinschmidt, line 922: what's the non-public part here?  `process_backtrace`?

kleinschmidt (approved):

test/runtests.jl
palday: yeah, applied to `catch_backtrace()`. The public method is on the stacktrace

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Legolas"
 uuid = "741b9549-f6ed-4911-9fbf-4a1c0c97f0cd"
 authors = ["Beacon Biosignals, Inc."]
-version = "0.5.24"
+version = "0.5.25"
```
palday:
```suggestion
version = "0.5.24"
```

